PR #180, opened 2014-12-01
Remove workaround for 32-bit msvc fenv_access bug

mwiebe:
Putting this up to trigger appveyor build which should show the issue.


mwiebe:
According to the connect site, the bug this is for is resolved as fixed.

PR #327, opened 2015-03-02
Work to make libdynd, CUDA, and Windows all live together peacefully

mwiebe:


mwiebe:
Still more problems, but I fixed a lot. Merging this for now.

PR #332, opened 2015-03-14
[WIP] Refactoring, especially of neighborhood

izaid:


include/dynd/kernels/apply.hpp
```diff
@@ -162,8 +152,7 @@ namespace nd {
       {
         if (val.get_type().get_type_id() == pointer_type_id) {
           m_val = val.f("dereference").as<T>();
-        }
-        else {
+        } else {
```
mwiebe: This is a bit annoying - the clang-format embedded in MSVC is behaving differently than other clang-format embeddings for this case.

include/dynd/strided_vals.hpp
```diff
@@ -8,284 +8,195 @@
...
+    static void incr(const char *arrmeta, const char **data, intptr_t *index)
+    {
+      ++index[0];
+      *data +=
+          reinterpret_cast<const fixed_dim_type_arrmeta *>(arrmeta)->stride;
+    }
+  };
+
+  template <typename T>
+  class fixed_dim<T, false> {
```
mwiebe: Couldn't the inner parameter be handled more cleanly by a recursive case? Something like:

```
template <typename T>
  class fixed_dim<fixed_dim<T>> {
 
```

izaid: What about if we want to support var_dim (not for neighborhood, but for apply or iteration)? I guess we could do that by simply inheriting from the fixed_dim<fixed_dim<T>> case?

mwiebe: Ok, so the reason to have this flag in there is to share implementation details between var_dim and fixed_dim, using the is_dim<...> metafunction.

izaid: The flag is because the innermost details need to be specialized. But, yes, fixed_dim[var_dim[T]] and fixed_dim[fixed_dim[T]] should behave the same way at the outermost fixed_dim.

include/dynd/strided_vals.hpp
```diff
@@ -8,284 +8,195 @@
...
-        intptr_t strides[N];
-    } m_mask;
-    intptr_t m_sizes[N];
-    const start_stop_t *m_start_stop;
-    intptr_t m_center_index[N];
+template <typename T>
+class fixed_dim : public detail::fixed_dim<T, !is_dim<T>::value> {
+  char m_arrmeta[fixed_dim::arrmeta_size];
+  const char *m_data;
+  const start_stop_t *m_start_stop;
```
mwiebe: What's the purpose of start_stop_t?

include/dynd/type.hpp
```diff
@@ -54,46 +59,51 @@ inline void *inc_to_alignment(void *ptr, size_t alignment) {
...
-char *iterdata_broadcasting_terminator_adv(iterdata_common *iterdata, intptr_t level, intptr_t i);
-char *iterdata_broadcasting_terminator_reset(iterdata_common *iterdata, char *data, intptr_t level);
+char *iterdata_broadcasting_terminator_incr(iterdata_common *iterdata,
+                                            intptr_t level);
+char *iterdata_broadcasting_terminator_adv(iterdata_common *iterdata,
+                                           intptr_t level, intptr_t i);
+char *iterdata_broadcasting_terminator_reset(iterdata_common *iterdata,
+                                             char *data, intptr_t level);
+
+// Forward declaration of strided_vals
```
mwiebe: Comment out of date here

include/dynd/strided_vals.hpp
izaid: It's a bit vestigal. We need a mechanism to support using fixed_dim in neighborhood. This tells which parts of the neighborhood are out of bounds. I'm trying to figure out a better way to do that...

include/dynd/type.hpp
izaid: Also true.

PR #334, opened 2015-03-22
[WIP] Arithmetic

izaid:


mrocklin:
Cool.  Glad to see pedestrian things like this happen.  

From an external perspective a fair amount of this seems repetitive; is there a way to generalize things? 

I ask because the [number of operators](https://docs.python.org/2/library/operator.html#mapping-operators-to-functions) is somewhat large and presumably one might want to change this code in the future.  Is it possible to define these operators and their computation/type interactions in a more compact way that would make future changes more pleasant?

include/dynd/types/type_id.hpp
```diff
@@ -622,4 +690,55 @@ struct is_type_bool<bool> {
   enum { value = true };
 };
 
-} // namespace dynd
+namespace detail {
+
+  template <typename T, int N>
+  class array_by_type_id;
```
mwiebe: Should have a docstring.

src/dynd/func/arithmetic.cpp
```diff
@@ -4,73 +4,73 @@
...
-#undef BUILTIN_TABLE
-#undef BUILTIN_ROW
+nd::arrfunc nd::plus::make()
+{
+  return functional::elwise(as_arrfunc<multidispatch_plus_ck>());
+}
+
+struct nd::plus nd::plus;
+
+nd::array nd::operator+(const nd::array &a0) { return nd::plus(a0); }
```
mwiebe: So nd::plus is both the type of the class and the instance, is that right? I'm used to that kind of thing being problematic, does it work fully consistently now?

src/dynd/kernels/arithmetic.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+  return ndt::type("(A0) -> R");
+}
+
+#undef MAKE_PAIRS
+
+#define MAKE_PAIRS(I, J)                                                       \
+  DYND_PP_JOIN_ELWISE(_MAKE_PAIRS, (, ), I, DYND_PP_REPEAT(J, DYND_PP_LEN(I)))
+#define _MAKE_PAIRS(I, J)                                                      \
+  DYND_PP_JOIN_ELWISE_1(MAKE_PAIR, (, ), DYND_PP_REPEAT(I, DYND_PP_LEN(J)), J)
```
mwiebe: I believe using identifiers like this with an _ at the beginning, is undefined, because the standard reserves such names for the compiler to do whatever it likes with. I think it would be a good practice to avoid using this kind of name.

mwiebe:
Looks like great progress to me. I think at some point, after the operators are filled in, we should take a look at the performance implications of using the big # type IDs by # type IDs table versus alternatives, and see if we can find a better approach.

@mrocklin I'm pretty sure it's possible to further reduce the amount of repeating boilerplate here. On the other hand I also think usually a good idea to start with a few instances of the code more fully written out, establishing and tweaking what the pattern is before generalizing.

izaid:
@mrocklin When I was implementing unary minus, I was thinking of you. ;)

src/dynd/func/arithmetic.cpp
izaid: It is. I'm actually using this pattern for all of our arrfuncs defined through declfunc. It does work well. If you want to specify the class, and not the instance, you can do it by using the keyword "struct".

izaid:
I do think there is a lot of repetition here, which is frustrating. I'd like to find a way to reduce that without making the code unreadable. For instance, I could use dynd's preprocessor more, but it makes it much harder to see what's going on. Let me see what I can do...

mrocklin:
Mostly I wanted to point to the [list of python operators](https://docs.python.org/2/library/operator.html#mapping-operators-to-functions) to flesh out a scope that would be useful from the Python user side.  

Happy to leave design bits to you all.

mwiebe:
Merged with Windows fixes from another PR

PR #340, opened 2015-03-30
[WIP] Data per call

izaid:


include/dynd/func/arrfunc.hpp
```diff
@@ -829,16 +823,19 @@ namespace nd {
           kwds.as_array(ndt::make_struct(self_tp->get_kwd_names(), kwd_tp),
                         available, missing);
 
+      // ...
+      char *data = new char[get()->size * arrfunc_type_data::data_size];
```
mwiebe: You can use std::unique_ptr<char[]> to hold this data, it would be nice to never have raw new/delete.

PR #399, opened 2015-05-14
Allow compilation with MinGW-w64.

insertinterestingnamehere:
This makes it so that compilation works with mingw-w64. With this, all tests pass on my windows machine. The 32 bit version of MinGW should also work, though I haven't tested it.


mwiebe:
Looks great! The g_in_fast_death_test_child change seems to have broken linux builds though.

insertinterestingnamehere:
Okay, this should fix that now. Are the test failures with cuda on Travis expected?

mwiebe:
Yes, the cuda support doesn't build because of arrfunc refactoring work that hasn't been reconciled with cuda yet.

PR #416, opened 2015-06-28
[WIP] Assignment

izaid:


include/dynd/kernels/multidispatch.hpp
```diff
@@ -349,6 +349,29 @@ namespace nd {
     template <int N0, int N1>
     struct new_multidispatch_by_type_id_kernel
         : base_virtual_kernel<new_multidispatch_by_type_id_kernel<N0, N1>> {
+      struct data {
```
mwiebe: From reading this, I think the i[2] is to allow multi-dispatching based on any two of the source types, or the dst + a source type, is that right? Might be worth a small comment to help ease reading the code.

include/dynd/types/type_id.hpp
```diff
@@ -313,106 +313,130 @@ struct type_id_of;
 
 template <typename T>
 struct type_id_of<const T> {
-  enum { value = type_id_of<T>::value };
+  static const type_id_t value = type_id_of<T>::value;
```
mwiebe: Yes, since dynd doesn't support older MSVCs any more, this is good.

izaid:
Yep -- it was also necessary to use value as a parameter to templates.

I think multidispatch is slowly converging, but it needs a lot of work still. I'm updating everything else and letting that drive how multidispatch works, it's all very experimental.

PR #452, opened 2015-07-14
Fix build errors from maybe-uninitialized warnings

insertinterestingnamehere:
This is the best version I was able to get for suppressing the maybe-uninitialized errors. Supposedly [according to this](http://stackoverflow.com/questions/5080848/disable-gcc-may-be-used-uninitialized-on-a-particular-variable), erroneous maybe-uninitialized and uninitialized errors aren't uncommon since the compiler is only able to use heuristics to provide a best guess as to whether or not they should be applied. I'm pretty sure the warning in expr_type.cpp was erroneous. As near as I can tell, the second warning in the float16 header is one where we may actually *want* to allow the cast to work with uninitialized data. A second pair of eyes would be greatly appreciated there though.

### DIFF
--- a/include/dynd/config.hpp
+++ b/include/dynd/config.hpp
@@ -52,6 +52,13 @@
 
 #define DYND_IGNORE_UNUSED(NAME) NAME __attribute__((unused))
 
+// Ignore erroneous maybe-uninitizlized
+// warnings on a given line or code block.
+#define DYND_IGNORE_MAYBE_UNINITIALIZED                                        \
+  _Pragma("GCC diagnostic ignored \"-Wmaybe-uninitialized\"")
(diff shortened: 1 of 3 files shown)


insertinterestingnamehere:
I opted to do the disabling/re-enabling in two separate macros rather than a single multi-line macro since the warnings were still showing (but not causing errors) when I used a multi-line macro to suppress them. That's probably a gcc bug, but this workaround should be good enough.

insertinterestingnamehere:
It would be good to get this one merged at some point.

mwiebe:
Looks reasonable to me

insertinterestingnamehere:
Thanks!

Issue #467, opened 2015-07-27
gcc 5.1 linker error


insertinterestingnamehere:
When compiling the tests with the latest master I'm getting a strange linker error.

    Linking CXX executable test_libdynd
    cd /home/owner/libdynd/gcc51_build/tests && /usr/bin/cmake -E cmake_link_script CMakeFiles/test_libdynd.dir/link.txt --verbose=1
[...]

PR #468, opened 2015-07-27
Include missing header file needed for tests.

insertinterestingnamehere:
This is a straightforward fix for a compile error that was showing up with gcc 5.1. See https://github.com/libdynd/libdynd/issues/467.


izaid:
Will merge when green.

insertinterestingnamehere:
Great! Thanks!

PR #475, opened 2015-07-30
Added numba to Python travis

izaid:


izaid:
Ended up going with conda.

PR #481, opened 2015-08-04
Try using the new container-based system on the Travis CI server.

insertinterestingnamehere:
This is mostly experimental to see if, and how well this works. I've disabled the Python build for now since it wasn't clear to me how to get libdynd in the right place to be used by dynd-python without sudo access.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,10 @@ matrix:
       addons:
         apt:
           sources:
+          - ubuntu-toolchain-r-test
           - kalakris-cmake
           packages:
+          - g++-4.7
           - cmake
           - valgrind
     - compiler: clang
@@ -42,8 +44,10 @@ matrix:
       addons:
         apt:
           sources:
+          - ubuntu-toolchain-r-test
           - kalakris-cmake
           packages:
+          - g++-4.7
           - cmake
           - libfftw3-dev
     - compiler: clang
@@ -51,8 +55,10 @@ matrix:
       addons:
         apt:
           sources:
+          - ubuntu-toolchain-r-test
           - kalakris-cmake
           packages:
+          - g++-4.7
           - cmake
           - libfftw3-dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,18 @@
 language: cpp
 
+sudo: false
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - gcc-4.7
+    - g++-4.7
+    - cmake
+    - valgrind
+    - libfftw3-dev
+
 compiler:
   - gcc
   - clang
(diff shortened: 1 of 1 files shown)


izaid:
Ha, this is awesome! We've been talking about doing this for a while, but have been held back because we didn't think we could install everything without sudo. Massive +1 if you can do it.

insertinterestingnamehere:
Thanks, I noticed that the SciPy devs got theirs working recently, so I'm feeling hopeful :)

insertinterestingnamehere:
Okay, except for the Python build, this is working. I've currently removed the Python build since it tends to fail all the time. If you'd like, I can figure out how to re-add it or re-add it as an allow-fail block so it can run and fail without making the indicator for the whole PR fail. Thoughts?

insertinterestingnamehere:
Okay, I haven't figured out how to get the Python build running yet. I'll try a few more things tomorrow. The latest version is my most recent attempt. It has all the others commented out to let me debug faster. Other than that, all the builds that use only C++ work fine.

izaid:
Okay, it's awesome that everything is working besides Python. That being said, I don't think we can drop the Python build -- it is too useful. We use it to determine if we broke the Python bindings and need to fix them, which happens frequently.

izaid:
@insertinterestingnamehere I merged the dynd-python PR and reran Travis, but it seems we're getting a segmentation fault at the test stage.

insertinterestingnamehere:
@izaid This looks like the segfault we were seeing with clang the other day. I'm baffled as to why it is only showing up on the new infrastructure and not on the current clang builds. I'll try enabling the verbose output and see if there are any differences in the flags.

izaid:
@insertinterestingnamehere Agreed. I think this segfault is now a major issue we need to solve. I'm also seeing a segfault on anaconda.org, although that is using gcc. I wonder if it is the same.

insertinterestingnamehere:
Hmm, rebuilding numpy with clang didn't fix it either. I'm not sure if this could show up with gcc. I'll try running the travis build here using gcc.

insertinterestingnamehere:
Okay, this all works with gcc. If you'd like I can re-enable the other builds and we can merge this for now.

izaid:
Wait, so do all the builds work?

insertinterestingnamehere:
Yep, I just commented the others out to speed up debugging the Python build. The Python build works with gcc but not with clang.

izaid:
Okay, we'll definitely merge this but I'd like to see if I can sort out the clang issue first.

izaid:
Oh, wait, are you saying that you want me to merge this now to help with sorting out the clang issue with Python? In that case I'll do it now.

insertinterestingnamehere:
Not really, I need to re-enable the other builds first. We could do that if
you'd like though.

izaid:
@insertinterestingnamehere I believe I just fixed the clang issue, at least on my local machine. I've merged a change into dynd-python. Do you want to try again now?

insertinterestingnamehere:
Fantastic! Yes, I'll switch this build back to clang and see if that fixes
it. Thanks!

insertinterestingnamehere:
Okay, the clang build works here now. I'll re-enable the other builds and get a similar PR up for dynd-python.

insertinterestingnamehere:
Alright, let's see if this works. Once the builds pass again this should be good to go.

insertinterestingnamehere:
Alright, this is done. Thanks!

izaid:
Fantastic work!

PR #471, opened 2015-07-28
[WIP] adding Windows dll support to libdynd

mwiebe:

PR #529, opened 2015-09-01
WIP: Visibility

insertinterestingnamehere:
This is a restart of https://github.com/libdynd/libdynd/pull/471 designed to allow full specification of private and public symbols in the libdynd library.

From my reading of the work already done by @mwiebe and various sources online, the visibility macros need to be applied to any exceptions we intend to raise, any classes we wish to explicitly hide or export (but not any of their corresponding methods), and any other objects or functions we wish to explicitly hide or export. They do not need to be applied to variables and functions declared with static visibility, variables, functions and classes declared in anonymous namespaces, and anything declared inline or as a template. The macros to specify what not to export will be most useful when hiding implementation details and making sure that the pydynd library does not leak symbols once it is made into a Python module.


insertinterestingnamehere:
Hahaha, okay, this has spectacularly many compiler errors at the moment, but the first pass to make all needed symbols visible is done. The next step in this PR is to go back and see if I can make the compiler happy that all the visibility declarations are consistent.

izaid:
This is lots of work. So far, it looks like major steps in the right direction!

insertinterestingnamehere:
Okay, so the next barrier to this appears to be that msvc raises a warning when an object from the standard library is exposed as a part of the interface for a dll. I'm trying to minimize this by more agressively inlining some functions and methods. From there, I think we should just suppress that particular warning, merge this, then continue working to remove the dependencies on standard library types in separate pull requests. Thoughts?

izaid:
Yep, I think that's absolutely the right approach. Suppress the MSVC warning, make this pass on all platforms, then feel free to merge it if it does.

insertinterestingnamehere:
Thanks for the input. This is working on msvc 2013 now. MinGW and msvc 2015 are still terribly averse to it though. :)

insertinterestingnamehere:
Merging. I'll follow up later today with more fixes and cleanup for MinGW and msvc 2015.

PR #530, opened 2015-09-01
Clang 3.7 compilation fixes

insertinterestingnamehere:
This removes a handful of unnecessary calls to std::move that cause warnings with the new clang/llvm. The new compiler warns if an unnecessary call to std::move prevents return value optimization. I've been aware of the issue for some time, but since Clang 3.7 is now officially released, adding this seemed sensible.


izaid:
Sounds good, merged.

PR #531, opened 2015-09-03
poking at the operator overloads to try and get MSVC 2015 building

mwiebe:
Filling in operators seems to just lead deeper down a rabbits hole, though.

### DIFF
--- a/include/dynd/config.hpp
+++ b/include/dynd/config.hpp
@@ -793,6 +793,21 @@ T floor(T value)
 
 } // namespace dynd
 
+#include <complex>
+
+// Add some complex std::common_type partial specializations
+namespace std {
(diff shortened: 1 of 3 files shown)


mwiebe:
It looks to me like something is broken about the SFINAE that should be happening within `has_make`, but not sure about the details yet.

izaid:
FYI, std::common_type is already overloaded for the complex types in complex.hpp

izaid:
@mwiebe I think this can be closed now.

insertinterestingnamehere:
Agreed. We may want to keep the int128 multiplication commit though.

izaid:
Closing this, because we've had MSVC 2015 working for some time.

PR #532, opened 2015-09-03
MSVC 2015 build fix attempt

insertinterestingnamehere:
Here's my attempt at fixing the issues with MSVC 2015. All I've managed to do thus far is change the nature of the errors the compiler spits out.

### DIFF
--- a/include/dynd/int128.hpp
+++ b/include/dynd/int128.hpp
@@ -178,6 +178,11 @@ public:
 
   //  DYND_CUDA_HOST_DEVICE inline int128 operator/(uint32_t rhs) const;
 
+  DYND_CUDA_HOST_DEVICE inline int128 &operator*(int128 DYND_UNUSED(rhs)) const
+  {
+    throw std::runtime_error("operator* is not implemented for int128");
+  }
+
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Okay, my goal here was to find a small workaround rather than adding a lot of extra code. I have it down to 3 errors, but am struggling to get it any further. I'll leave it open for reference while we work on getting this fixed. Mark's larger refactoring in https://github.com/libdynd/libdynd/pull/531 is probably the way to go.

insertinterestingnamehere:
I now have a msvc 2013 environment working, so at least this is no longer causing problems for the visibility fixes.

mwiebe:
Let's merge this then, and continue with 2015 separately.

insertinterestingnamehere:
Ah, sorry I wasn't clear, getting 2013 working on my end stemmed from getting the environment for it working properly and was independent of these changes. There's little or no net gain from this PR for the repository in general other than a small reduction in the number of msvc2015 errors we're seeing. Aside from that, this is just a handful of unnecessary special cases. I'll open a PR that reverts it.

PR #536, opened 2015-09-09
WIP: Visibility

insertinterestingnamehere:
Further cleanup of visibility specifications to cut down on warnings.


insertinterestingnamehere:
This is good to go. The rest of the changes ended up being relatively straightforward. I did make it so that `dynd::dynd_exception` no longer subclasses from `std::exception` so that it can pass across dll boundaries. If there are no objections, I'd like to merge this soon and continue with more modifications to the build system in a separate PR.

insertinterestingnamehere:
https://github.com/libdynd/dynd-python/pull/354 fixes the dynd-python error in the Python build.

include/dynd/exceptions.hpp
```diff
@@ -23,7 +22,7 @@ namespace nd {
   class array;
 } // namespace nd
 
-class DYND_API dynd_exception : public std::exception {
+class DYND_API dynd_exception {
```
mwiebe: One suggestion I have here is to add a comment describing why it's not a subclass of std::exception, as that seems like a pretty obvious question to ask when browsing this code.

insertinterestingnamehere: Good idea, I'll add that in a follow up PR.

PR #539, opened 2015-09-11
Parallel Compilation

insertinterestingnamehere:
This enables parallel compilation in the Travis CI build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,12 +23,12 @@ matrix:
     - export CC="gcc-4.8"
     - export CXX="g++-4.8"
     script:
-      - make || exit 1
+      - make -j5 || exit 1
       - valgrind --tool=memcheck --leak-check=yes --error-exitcode=123 ./tests/test_libdynd
   - compiler: clang
     env: VALGRIND DYND_FFTW=OFF
     script:
-      - make || exit 1
+      - make -j5 || exit 1
       - valgrind --tool=memcheck --leak-check=yes --error-exitcode=123 ./tests/test_libdynd
   - compiler: clang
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Okay, enabling parallel compilation for msvc made appveyor terribly slow, so that'll just have to be done via the arguments to cmake for now.

insertinterestingnamehere:
I just enabled parallel building for the conda build as well. I haven't taken the time to make the number of threads an option. I just set it to default at 5 so that it will run well on the two-core Travis CI machines. @izaid any thoughts on that?

izaid:
Wow, is it actually twice as fast with clang? That's impressive, feel free to merge so long as nothing is terribly slow.

insertinterestingnamehere:
The only slowdown was because appveyor couldn't handle running msvc with the -MP flag. On my windows machine doing that speeds up compilation dramatically, but the single-core appveyor bot wouldn't have anything of it, so I removed that change. Merging now.

mwiebe:
I wonder if it's worth using http://www.cmake.org/cmake/help/v3.0/module/ProcessorCount.html to set the default?

insertinterestingnamehere:
Yep, we could definitely do some kind of dynamic processor detection in the conda build. There really isn't any point in doing that in the Travis build. On the other hand, the conda build calls cmake and then make, so setting the number of cores automatically may need to be done in shell. I haven't found any way to have cmake set the default number of threads for a makefile yet.

insertinterestingnamehere:
Okay, I took a shot at it in https://github.com/libdynd/libdynd/pull/540.

PR #540, opened 2015-09-11
Dynamically detect the number of CPU cores in conda build.

insertinterestingnamehere:


insertinterestingnamehere:
Only for Linux and Mac for now.

We could do the same thing in the conda build for windows by adding `-DCMAKE_C_FLAGS="-MP" -DCMAKE_CXX_FLAGS="-MP"` to the cmake configuration step. I'm reluctant to do that by default in CMakeLists.txt for Windows right now though because it makes AppVeyor take around twice as long as before.

groutr:
Would the `CPU_COUNT` environment variable that conda-build determines work here?

insertinterestingnamehere:
@groutr Thanks! Yes, it would. I'll see if that works on our CI systems.

insertinterestingnamehere:
@izaid feel free to merge this one if you'd like. It doesn't appear to have had any impact on our CI tests, but it will speed things up for any builds done locally with conda.

izaid:
Great, thanks!

PR #541, opened 2015-09-11
Further Symbol Visibility fixes

insertinterestingnamehere:


insertinterestingnamehere:
My goal here is to make it so that we can (finally) build a usable dll and then make that the default behavior. There are still many linker errors that come when trying to build a dll, but I'll be pushing progress here as it comes.

insertinterestingnamehere:
@mwiebe @izaid It looks like I need to use a cmake templating mechanism to make it so that if we are installing a shared library the preprocessor defines give `__declspec(dllimport)` in all the appropriate places and if we are installing a static library all the dllimports are not present. Can either of you confirm that? It seems weird to be changing one of the main macros like that, but that's the only way I see for supporting using both static and dynamic libraries.
As things are, this version works for building a shared lib (other than the fact that the libdynd.dll needs to be copied into the folder for the tests to get them to run properly), but it fails badly for a static build. On the other hand, if I make it so that the dllimport macro isn't present for client modules, everything works fine.

insertinterestingnamehere:
Okay, I think I see where to go from here. [It appears](http://stackoverflow.com/questions/4489441/why-when-is-declspec-dllimport-not-needed) this is only a problem for data and objects exported from the dll. Those, I should be able to fix by declaring things like `dynd::ndt::type::instances`, `dynd::eval::default_eval_context`, `dynd::ndt::detail::builtin_kinds`, `dynd::ndt::detail::builtin_data_alignments`, `dynd::ndt::type::instances`, `dynd::ndt::detail::builtin_data_sizes`, `dynd::string_encoding_char_size_table`, `dynd::nd::sin`, `dynd::nd::random::uniform`, `dynd::nd::mean`, etc. in their corresponding header files as static rather than having them be declared extern in the headers and defined in the cpp files.

insertinterestingnamehere:
Okay, this is good locally. We'll see if appveyor agrees.

insertinterestingnamehere:
Okay, this is passing. I'm still hunting out symbols that aren't used by the tests but still should be exported. This can be merged as it is though.

izaid:
This is a lot of solid work, thanks a lot!

insertinterestingnamehere:
Thanks for the help!

PR #542, opened 2015-09-12
[WIP] LLVM-based Clang plugin

izaid:
This is a major new feature that gives DyND the ability to automatically store the LLVM IR of all its callables when compiled with clang. Ultimately, we will bring this to GCC (via dragonegg) and Windows (via clang on Windows, if there is nothing that works for MSVC) as well.


seibert:
This sounds very interesting...  Can you explain a little more what the motivation is for this feature?

PR #551, opened 2015-09-17
Properly export const arrays from dll.

insertinterestingnamehere:


insertinterestingnamehere:
With this last change, the tests all compile and pass for both static and dynamic libs on both msvc 2013 and mingw-w64 (gcc 5.2).

insertinterestingnamehere:
No news for cuda on windows. I just gave that build a shot and it is definitely broken, though maybe I'm not telling cmake to use nvcc properly. We'll have to revisit that at some point, but now probably isn't the time.

PR #556, opened 2015-09-18
Removed array_rw.

insertinterestingnamehere:


izaid:
Closing this, because I removed array_rw in a different PR. There is still the question of what the right flags are, but at least everything now goes through the array constructor.

PR #557, opened 2015-09-19
[WIP] Custom types

izaid:
This PR adds dynamically registerable types to DyND.


insertinterestingnamehere:
This is fantastic! It's great that we can start doing stuff like this now!

PR #558, opened 2015-09-20
Add a function for parsing generic datashape type constructor arguments

mwiebe:
This is exposed as its own function, and tested as such. Has not been hooked up elsewhere in the type string parser yet.


izaid:
@mwiebe This seems to be broken on every compiler that's *not* MSVC -- the opposite of usual. :)

If you could fix this so we can merge, that'd be great!

PR #561, opened 2015-09-21
Turn zeros and ones into kernels

cpcloud:


izaid:
Want to look through this, @insertinterestingnamehere?

cpcloud:
I see the problem

insertinterestingnamehere:
Hmm. That's a really odd test failure. Looking into it more now.

izaid:
@insertinterestingnamehere -- Actually, I was just thinking it'd be good for you to review the code. I don't think you need to fix any failures.

insertinterestingnamehere:
Good idea. Thanks. I've been wanting to read through more examples of going from function to kernel to arrfunc.

izaid:
@insertinterestingnamehere If you can do this as a formal code review, that's be great. Leave comments for anything that needs to be fixed.

insertinterestingnamehere:
Will do.

src/dynd/func/fill.cpp
```diff
@@ -0,0 +1,109 @@
...
+        uint32_type_id,
+        uint64_type_id,
+        uint128_type_id,
+        float16_type_id,
+        float32_type_id,
+        float64_type_id,
+        float128_type_id,
+        complex_float32_type_id,
+        complex_float64_type_id
+    > type_ids;
```
insertinterestingnamehere: We should be able to just use some of our predefined typedefs here. [aritmetic_type_ids](https://github.com/libdynd/libdynd/blob/master/include/dynd/types/type_id.hpp#L207) is the same as what you define here.

src/dynd/func/fill.cpp
```diff
@@ -0,0 +1,109 @@
...
+         0);
+}
+
+DYND_API struct nd::zeros nd::zeros;
+
+DYND_API nd::callable nd::ones::children[DYND_TYPE_ID_MAX + 1];
+
+DYND_API nd::callable nd::ones::make()
+{
+    typedef type_id_sequence<
```
insertinterestingnamehere: Same as above, use one of the prebuilt lists for this.

insertinterestingnamehere:
I'm fine with these changes once the tests pass. Looking forward, I do have some questions about these routines though:
What is the benefit of using an arrfunc for these allocation routines as opposed to a convenience function?
In the long run, should we expect looping through an array for initialization to be optimized by the compiler, or should we work towards doing this sort of initialization via a call to `memset`?
Is there any reason we chose to make two nearly identical arrfuncs for filling an array with values, and is there a straightforward way to remove the repetition?

izaid:
@cpcloud What's the status on this? Would you mind finishing it up?

PR #562, opened 2015-09-22
Operators

insertinterestingnamehere:


include/dynd/kernels/arithmetic.hpp
```diff
@@ -7,42 +7,66 @@
 namespace dynd {
 namespace nd {
 
-  template <type_id_t I0>
-  struct plus_kernel : base_kernel<plus_kernel<I0>, 1> {
+  template <type_id_t I0, typename func_type, func_type f>
+  struct unary_kernel : base_kernel<unary_kernel<I0, func_type, f>, 1> {
     typedef typename type_of<I0>::type A0;
```
izaid: A few changes I would suggest here. We should rename I0, I1, ... to Src0TypeID, Src1TypeID, ... (this was introduced to the other kernels later on after the arithmetic_kernels were written). We also should write template parameters AsFirstLetterCapitalized. @mwiebe and I decided on this at one point and we're trying to be consistent now.

include/dynd/kernels/arithmetic.hpp
```diff
@@ -7,42 +7,66 @@
...
-  template <type_id_t I0>
-  struct minus_kernel : base_kernel<minus_kernel<I0>, 1> {
-    typedef typename type_of<I0>::type A0;
-    typedef decltype(-std::declval<A0>()) R;
+#define DYND_DeclUnaryOp(OP, NAME)                                                                        \
+  template <typename T>                                                                                   \
+  inline decltype(OP std::declval<T>()) inline_ ## NAME(T val) { return OP val; }                         \
+  template <type_id_t I0>                                                                                 \
+  struct NAME ## _kernel : unary_kernel<I0, decltype(&inline_ ## NAME<typename type_of<I0>::type>),       \
+                                        &inline_ ## NAME<typename type_of<I0>::type>> {};                 \
```
izaid: So, in terms of code duplication this is obviously better. I definitely the idea of having kernels being templated on the operation and want to keep that. A few thoughts:

1) Should we use macros here? It does make the code more unreadable.

2) I think we might be able to reuse the kernels in apply_kernels.hpp for this. We should check that out.

insertinterestingnamehere: I prefer the macros primarily because they let the reader know that the *exact* same thing is being done several times over, but that's just me. I can see why we would prefer not to have them though, so we can get rid of them if you'd like.

Thanks for pointing out the kernels there. It makes a lot of sense to use those instead.

insertinterestingnamehere:
I'm a little worried about the msvc failure. It looks like it's complaining about using an explicit template instantiation as a template parameter. Hopefully using the kernels in `kernels/apply.hpp` will help in working around the issues here.

insertinterestingnamehere:
I'm curious if we should try rolling comparisons in with these operators as well. Since we can have the compiler determine the return-type via templates, I don't know if there's a compelling reason to use different code for them.

include/dynd/kernels/arithmetic.hpp
insertinterestingnamehere: That makes a lot of sense. I'll go ahead and make that change.

insertinterestingnamehere:
Okay, I believe I've finally found a fix to get this to work with MSVC. Hopefully the CI builds agree.

PR #569, opened 2015-09-27
Operators

insertinterestingnamehere:
This is a little more work toward simplifying the operator code. This enables `operator+=` by restricting the number of types the in-place operators apply to. From here I plan on expanding the number of operators available, adding dummy functions for all the missing operators, and then re-enabling the full list of numeric types for all the different operators.


izaid:
Looks great, merging!

PR #576, opened 2015-09-30
Document that MSVC 2015 is now required

insertinterestingnamehere:
This documents that we are now requiring MSVC 2015. There are a few MSVC 2013 specific workarounds in config.hpp that we should probably remove in a few more weeks.


izaid:
+1 on the updating docs

PR #577, opened 2015-09-30
WIP: Removed implicit conversions to DyND arrays

insertinterestingnamehere:
This is still far from finished, but I'll post it for review anyway. Thus far I have managed to make it so that the main libdynd library compiles. Now I'm working on the tests.

### DIFF
--- a/include/dynd/array.hpp
+++ b/include/dynd/array.hpp
@@ -80,31 +80,31 @@ namespace nd {
      * Constructs a zero-dimensional scalar from a C++ scalar.
      *
      */
-    array(bool1 value);
-    array(bool value);
-    array(signed char value);
-    array(short value);
-    array(int value);
-    array(long value);
-    array(long long value);
-    array(const int128 &value);
-    array(unsigned char value);
-    array(unsigned short value);
-    array(unsigned int value);
-    array(unsigned long value);
-    array(unsigned long long value);
-    array(const uint128 &value);
-    array(float16 value);
-    array(float value);
-    array(double value);
-    array(const float128 &value);
-    array(complex<float> value);
-    array(complex<double> value);
-    array(std::complex<float> value);
-    array(std::complex<double> value);
-    array(const std::string &value);
+    explicit array(bool1 value);
+    explicit array(bool value);
+    explicit array(signed char value);
+    explicit array(short value);
+    explicit array(int value);
+    explicit array(long value);
+    explicit array(long long value);
+    explicit array(const int128 &value);
+    explicit array(unsigned char value);
+    explicit array(unsigned short value);
+    explicit array(unsigned int value);
+    explicit array(unsigned long value);
+    explicit array(unsigned long long value);
+    explicit array(const uint128 &value);
+    explicit array(float16 value);
+    explicit array(float value);
+    explicit array(double value);
+    explicit array(const float128 &value);
+    explicit array(complex<float> value);
+    explicit array(complex<double> value);
+    explicit array(std::complex<float> value);
+    explicit array(std::complex<double> value);
+    explicit array(const std::string &value);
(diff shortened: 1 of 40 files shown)


izaid:
I just read through it. I'm a bit concerned that we need to wrap many things nd::array now -- I think that makes things harder to use. Are we just missing explicit constructors for those cases?

insertinterestingnamehere:
Yep, a better solution is to choose which things we want to let have a more general interface and declare the needed overloads. I can do a pass through the code to define the extra overloads later in this PR if you'd like. The unfortunate downside of the current approach is that something like this could happen to one of our users...

    double f(vector<double> a) { ... }
    double f(nd::array a) { ... }
    ...
    ... f({1., 2.}); // Error on ambiguous call here

I'm working on spinning up an actual example of that happening. The problem is that `std::vector` already allows implicit conversion from an initializer list, so if both headers are present the call becomes ambiguous.

insertinterestingnamehere:
On the other hand, the argument could be made that if applicable overloads exist for both `std::vector` and `dynd::nd::array`, the user's meaning in calling a function with an initializer list really is ambiguous. I'm inclined toward not allowing the implicit conversion in general, but it may be something we still want to allow with initializer lists.

izaid:
I think you have a good point for the situation where it is ambiguous, but the compiler would complain there anyway. My main concern is for functions that just take an nd::array, like callable_type::make. That should be able to accept an initializer_list.

insertinterestingnamehere:
I can look into ways to allow the conversion to take place for each type's `make` routine. Hopefully there is a way to do it via subclassing rather than by defining extra overloads everywhere. I'm also planning on making sure callables can still accept scalars and initializer lists. I'm also going to refine the templates I put in for the operator overloads here so that they are more specific to the things we want to allow. Where else do we want to allow conversions like this to take place?

izaid:
I think any argument that is an nd::array that is not ambiguously overloaded take an initializer list. So void func(const nd::array &a) should do it, but void func(const nd::array &a) with func(const std::vector<T> &v) is allowed to complain.

I'm actually confused why this doesn't still work, even with explicit.

insertinterestingnamehere:
Here's a minimum example showing the kind of strange errors we're trying to avoid:

    #include <vector>
    #include <iostream>
    #include <dynd/array.hpp>
    int f(dynd::nd::array a) { return a(0).as<int>(); }
    int f(std::vector<int> a) { return a[0]; }
    int main() { std::cout << f({1, 2, 3}); }

Aside from that, one of my main motivations for looking at this now is that using fewer implicit conversions (for scalars in particular) should make the template errors from failed operator overloading easier to follow.

insertinterestingnamehere:
Weird errors aside, from an API standpoint I can see why we would want to allow conversions from initializer lists. The implicit conversions to and from scalars worry me a lot more. For example, the following takes what one would expect to be a compilation error and turns it into a runtime error:

    #include <vector>
    #include <iostream>
    #include <dynd/array.hpp>
    int f(dynd::nd::array a) {
      return a(0).as<int>();
    }
    
    int main() {
      std::cout << f(3.);
    }

Behavior like that may make sense for dynd callables, but doing that to user-defined functions seems extreme.

Anyway, discussion as to what is the ideal way to resolve this is welcome. I've opened this mostly to show one direction we can start moving toward.

insertinterestingnamehere:
Quick update here. I just finished digging into some C++ oddities we had here. From there I've shifted toward getting the tests to work. I think the best way forward is to get all the tests working again and then focus on getting more of the needed overloads working while simultaneously removing the explicit conversions to arrays that had to be inserted to get everything working.

insertinterestingnamehere:
Closing this for now. I'll leave the branch up for a while though in case we end up needing any of it.

PR #579, opened 2015-10-01
Operators

insertinterestingnamehere:
This is a set of further simplifications for the operator code that I wrote a few days ago while working to enable more operators. Since the work on removing implicit conversions is slowing my ability to add new operators, it makes sense to put these changes in now rather than waiting any longer.


izaid:
Looks good, merged!

insertinterestingnamehere:
Thanks!

PR #581, opened 2015-10-05
Gcc version bump

insertinterestingnamehere:
This enables c++14 and drops support for gcc 4.8.


insertinterestingnamehere:
Okay, this is working now. The ubuntu packages for clang 3.4 and clang 3.5 don't work with the gcc 4.9 libstdc++, so on the linux machines I switched to installing clang 3.5 from the corresponding llvm package repositories (the 3.4 repository isn't whitelisted for the travis ci servers, so I went with 3.5).

I also made it so our valgrind builds run with gcc 5 and clang 3.7 to diversify the compiler versions present in our build.

izaid:
Okay, wow, that's highly annoying. It's not great that this is the case, but I don't see what alternatives we have so I'm going ahead and merging this. Thanks for the hard work at sorting this out.

PR #583, opened 2015-10-06
[WIP] Clang on Travis

izaid:


izaid:
@insertinterestingnamehere I was able to make Clang 3.4 and later work on Travis CI. We need to use the -std=c++1y flag. Following that, Google Benchmark still causes problems. That is fixable by removing two unnecessary "#include <cstddef>".

I think we should go this route, as I don't want to make it impossible to use Ubuntu's Clang. We'll turn the -std=c++1y flag into -std=c++14 when that is no longer necessary.

insertinterestingnamehere:
@izaid Fantastic! I'm glad this is working now. That sounds good to me.

PR #584, opened 2015-10-06
Updated Google Benchmark

izaid:


insertinterestingnamehere:
Now that google benchmark and google test are on github too, why don't we just include them as git submodules?

izaid:
We should definitely do that.

izaid:
@insertinterestingnamehere Good suggestion, just did this for benchmark. The only downside is people need to specify --recursive when cloning.

insertinterestingnamehere:
@izaid Cool. I'm glad we have that working now. The recursive clone is unfortunate though.

PR #585, opened 2015-10-07
[WIP] JSON discovery tool

izaid:
This adds a new function that determines the type of the corresponding DyND array from a JSON blob.


izaid:
Merging progress so far, will need to come back to fix up a few things.

PR #586, opened 2015-10-07
Remove erroneous overloads of copy assignment operators

insertinterestingnamehere:
This is one code-cleanup item from https://github.com/libdynd/libdynd/pull/577 that we should probably keep. It removes an unnecessary extra template that was introduced to avoid errors with implicit conversions, but actually doesn't do anything.


izaid:
Thanks!

PR #587, opened 2015-10-08
Operators

insertinterestingnamehere:
More work toward expanding the types to which the arithmetic operators apply.

### DIFF
--- a/include/dynd/func/arithmetic.hpp
+++ b/include/dynd/func/arithmetic.hpp
@@ -14,11 +14,11 @@
 namespace dynd {
 namespace nd {
 
-  template <typename F, template <type_id_t...> class K, int N, typename TypeIDs>
+  template <typename F, template <type_id_t...> class K, int N, typename TypeIDSequence>
   struct arithmetic_operator;
 
-  template <typename FuncType, template <type_id_t> class KernelType, typename TypeIDs>
-  struct arithmetic_operator<FuncType, KernelType, 1, TypeIDs> : declfunc<FuncType> {
+  template <typename FuncType, template <type_id_t> class KernelType, typename TypeIDSequence>
+  struct arithmetic_operator<FuncType, KernelType, 1, TypeIDSequence> : declfunc<FuncType> {
     static callable children[DYND_TYPE_ID_MAX + 1];
 
     static callable &overload(const ndt::type &src0_tp)
@@ -33,7 +33,7 @@ namespace nd {
           functional::call<FuncType>(ndt::type("(Any) -> Any"));
 
       for (const std::pair<const type_id_t, callable> &pair :
-           callable::make_all<KernelType, TypeIDs>(0)) {
+           callable::make_all<KernelType, TypeIDSequence>(0)) {
         children[pair.first] = pair.second;
       }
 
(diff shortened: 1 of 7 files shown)


insertinterestingnamehere:
Let's not merge this just yet. We need to finish planning out how to make all these changes, and the expression SFINAE here may or may not be something we want to use to work toward getting everything working.

include/dynd/func/arithmetic.hpp
```diff
@@ -14,11 +14,11 @@
 namespace dynd {
 namespace nd {
 
-  template <typename F, template <type_id_t...> class K, int N>
+  template <typename F, template <type_id_t...> class K, int N, typename TypeIDs>
   struct arithmetic_operator;
```
izaid: Adding the TypeIDs as a template parameter is a good step for sure. A trivial comment is that I typically have called such a typename TypeIDSequence throughout DyND, so we should harmonize that.

A less trivial comment is I'd like to hear your opinions on this arithmetic_operator template. This template probably should be a more general thing, as we have a similar comparision_operator template and so on. I'm not sure if arithmetic_operator was a good idea or not, it was a bit experimental when I wrote it.

include/dynd/kernels/arithmetic.hpp
```diff
@@ -8,35 +8,94 @@
 namespace dynd {
 namespace nd {
 
+namespace detail{
+  template<typename>
+  struct sfinae_true : std::true_type{};
+}
```
izaid: Things like sfinae_true should go into config.hpp as they are more general. But, we should be really careful about adding things like that.

sfinae_true seems like something you need, but there is a true_t on config.hpp which does the same thing. Probably best to use that.

include/dynd/kernels/arithmetic.hpp
```diff
@@ -430,41 +486,69 @@ namespace nd {
...
+    static type make()                                                 \
+    {                                                                  \
+      std::map<std::string, ndt::type> tp_vars;                        \
+      tp_vars["A0"] = ndt::type::make<A0>();                           \
+      tp_vars["R"] = ndt::type::make<R>();                             \
+                                                                       \
+      return ndt::substitute(ndt::type("(A0) -> R"), tp_vars, true);   \
+    }                                                                  \
+  };                                                                   \
+
```
izaid: I'm the one who originally write the make function here, so I'm just criticizing myself. We shouldn't use ndt::substitute in this functions. At the time, we lacked tools to easily construct a callable_type, but we have those now. One should use something like `callable_type::make(ndt::type::make<R>(), {ndt::type::make<A0>()})`.

include/dynd/kernels/arithmetic.hpp
```diff
@@ -430,41 +486,69 @@ namespace nd {
...
+    static type make()                                                      \
+    {                                                                       \
+      std::map<std::string, ndt::type> tp_vars;                             \
+      tp_vars["A0"] = ndt::type::make<A0>();                                \
+      tp_vars["A1"] = ndt::type::make<A1>();                                \
+      tp_vars["R"] = ndt::type::make<R>();                                  \
+                                                                            \
+      return ndt::substitute(ndt::type("(A0, A1) -> R"), tp_vars, true);    \
+    }                                                                       \
+  };                                                                        \
```
izaid: Same as above.

insertinterestingnamehere:
Okay, I've addressed the comments here, and, as per our discussion this morning, opted not to go the expression SFINAE route. I didn't see any clear way to generalize the `DYND_HAS` macro to apply to the expressions here, but that's fine since we aren't using the expression SFINAE anymore. I'll deal with figuring out whether or not we want the arithmetic kernel infrastructure when I merge the comparison operators and the arithmetic operators into the same file. This should be good to go.

izaid:
Great, merged!

PR #591, opened 2015-10-11
Restructuring available lists of type ID's

insertinterestingnamehere:
Proposed restructuring of the existing lists of types.

### DIFF
--- a/include/dynd/float128.hpp
+++ b/include/dynd/float128.hpp
@@ -270,6 +270,10 @@ public:
 
   */
 
+  float128 operator+() const {
+    return *this;
+  }
+
   float128 operator-() const
   {
     return float128(-static_cast<double>(*this));
--- a/include/dynd/float16.hpp
+++ b/include/dynd/float16.hpp
@@ -274,6 +274,10 @@ public:
 
   DYND_CUDA_HOST_DEVICE friend float16 float16_from_bits(uint16_t bits);
 
+  float16 operator+() const {
+    return *this;
+  }
+
   float16 operator-() const
   {
     return float16(-static_cast<float>(*this));
--- a/include/dynd/func/arithmetic.hpp
+++ b/include/dynd/func/arithmetic.hpp
@@ -73,8 +73,8 @@ namespace nd {
     }                                                                                                \
   } NAME;                                                                                            \
 
-  DYND_DEF_UNARY_OP_CALLABLE(plus, numeric_type_ids)
-  DYND_DEF_UNARY_OP_CALLABLE(minus, numeric_type_ids)
+  DYND_DEF_UNARY_OP_CALLABLE(plus, complex_type_ids)
+  DYND_DEF_UNARY_OP_CALLABLE(minus, complex_type_ids)
 
 #undef DYND_DEF_UNARY_OP_CALLABLE
 
--- a/include/dynd/func/arithmetic.hpp
+++ b/include/dynd/func/arithmetic.hpp
@@ -73,8 +73,8 @@ namespace nd {
     }                                                                                                \
   } NAME;                                                                                            \
 
-  DYND_DEF_UNARY_OP_CALLABLE(plus, complex_type_ids)
-  DYND_DEF_UNARY_OP_CALLABLE(minus, complex_type_ids)
+  DYND_DEF_UNARY_OP_CALLABLE(plus, arithmetic_type_ids)
+  DYND_DEF_UNARY_OP_CALLABLE(minus, arithmetic_type_ids)
 
 #undef DYND_DEF_UNARY_OP_CALLABLE
 
--- a/include/dynd/types/type_id.hpp
+++ b/include/dynd/types/type_id.hpp
@@ -199,14 +199,13 @@ enum type_id_t {
 template <type_id_t... I>
 using type_id_sequence = integer_sequence<type_id_t, I...>;
 
-typedef type_id_sequence<int8_type_id, int16_type_id, int32_type_id, int64_type_id, int128_type_id> signed_int_type_ids;
-typedef type_id_sequence<bool_type_id, uint8_type_id, uint16_type_id, uint32_type_id, uint64_type_id, uint128_type_id> unsigned_int_type_ids;
-typedef type_id_sequence<float16_type_id, float32_type_id, float64_type_id, float128_type_id> floating_type_ids;
-typedef type_id_sequence<complex_float32_type_id, complex_float64_type_id> complex_floating_type_ids;
+typedef type_id_sequence<int8_type_id, int16_type_id, int32_type_id, int64_type_id, int128_type_id> int_type_ids;
+typedef type_id_sequence<bool_type_id, uint8_type_id, uint16_type_id, uint32_type_id, uint64_type_id, uint128_type_id> uint_type_ids;
+typedef type_id_sequence<float16_type_id, float32_type_id, float64_type_id, float128_type_id> float_type_ids;
+typedef type_id_sequence<complex_float32_type_id, complex_float64_type_id> complex_type_ids;
 
-typedef join<signed_int_type_ids, unsigned_int_type_ids>::type integral_type_ids;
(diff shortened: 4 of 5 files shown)


insertinterestingnamehere:
I'm still not a fan of my naming choice for the complex_floating types vs the complex types. Feedback welcome, particularly on ideas for better naming conventions.

izaid:
Yeah, I'm not a fan of your naming choice either -- but I'm also not a fan of *my* naming choice or @mwiebe's naming choice before that. :)

Perhaps it's time to fix this? Here are some thoughts.

An obvious guiding principle is that the names for type IDs, and their corresponding type ID sequences, should be clear and short. I'd also add that, in the absence of a reason not to, we should follow C++ terminology where possible. If one sticks with that, and accepts that our complex types are really first-class types, you quickly arrive at the following list.

- We drop the "signed" from "signed_int", and likewise the "s" from "sint".
- We use "uint" to mean unsigned integers.
- We don't use "real", but use "float" instead.
- We use "arithmetic" to refer to bool, any int, and uint, any float, or any complex. This is analogous to `std::is_arithmetic`, except we added complex.
- We use "fundamental" to refer to anything in "arithmetic" and also void.
- We repurpose "builtin" to mean any type in DyND that has a type ID in the type_id_t enum.

@mwiebe @insertinterestingnamehere Is this good or bad?

insertinterestingnamehere:
@izaid that sounds file to me. When in C++ we can do as C++ does. Here are some issues that I still haven't figured out:

- For the operator work, assuming we decide to disallow comparisons with complex numbers, we'll need a list of types that contains integer and floating point types but not our complex types. I opted to call that "real", but we could call it something like "ordered."
- We don't actually use  the fundamental or builtin type lists for anything right now. Should we include them for possible use later or wait until a use case arises? Do we have any sort of general policy about including stuff like that? Do we intend for these lists to be internal or not?
- This one's a little pedantic, but, internally, we'll be using "aritmetic" to indicate that a type has algebraic field-like operations defined. That's a bit of a misnomer since you can do arithmetic on many other things. It's a rare enough case that we'd be fine using names like "additive_group_arithmetic", "ring_arithmetic", or "integral_domain_arithmetic" to specify that sort of thing if the need arises. Really this concern has more bearing on user-defined types anyway.

I'll put up a modified proposal that incorporates the C++-like naming while still leaving out the fundamental and builtin lists and including a list of "ordered" types that specifies both integral and floating point types.

izaid:
@insertinterestingnamehere Looks good, feel free to merge.

insertinterestingnamehere:
Great! Will do. Thanks!

izaid:
- I don't think we need a list of all the built-in type IDs, but we do need the maximum built-in type ID declared somewhere. I'd keep the fundamental type ID list, seems important. We have no general policy.

- Yeah, looking at the arithmetic type ID sequence I'm also not totally happy with the name. Same with ordered, actually. Maybe we should just drop these and use typename dynd::join<...>::type a lot?

A lot of this is semantics, but these are the sort of thing that we do need to sort out. So far, we've been kicking the can down the road a lot, and I'm all for that stopping.

insertinterestingnamehere:
Okay, the maximum built-in type ID is still declared, so that should be fine. I'll go ahead and add the fundamental type id list back in. Using joins all over for now would work in the immediate future, though that doesn't take us much further than just listing things where they are used.

insertinterestingnamehere:
Here are a few potentially useful observations here, in no particular order:

- ordering, mathematically, is a property that is fundamentally independent of any other abstract-algebraic structure, so we don't really need to connect the two in our naming choices.

- NumPy uses the word "scalar" for types that permit all arithmetic operations. That's not a bad naming convention, except that it conflicts somewhat with the c++ definition of scalar which includes pointer, pointer to member, and std::nullptr types.

- Really the minimal algebraic structure we need for division to be applicable is a Euclidean domain with a specifically chosen valuation function. Unfortunately, for the mod operator to be at all useful, it is important that the type represent a Euclidean domain that is *not* an algebraic field. Thankfully, the only types to which that applies thus far are integers, so we may not need to make as much of a distinction yet.

- One possible way to handle this would be to list types according to which operations are applicable. That seems a little too verbose to expose to users though.

- Following C++ semantics, even with the over-generic-ness of "arithmetic" may still be the best way to avoid as much confusion as possible.

mwiebe:
About ordering, take a look at http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4367.html if you haven't seen it. For data structure and algorithm purposes, you often want ordering to work a certain way (e.g. NaN == NaN being true, `complex<double>` to have an ordering). It probably doesn't affect this discussion too much, but what is currently called `sorting_less` in the dynd code should maybe follow the better naming expressed in that paper.

insertinterestingnamehere:
@mwiebe Thanks for sharing that. It does help some.
I'd definitely welcome further discussion on whether our complex numbers should allow comparisons or not. Mathematically, I see compelling reasons why it shouldn't be allowed (the ordering induces a topology that is wholly incompatible with the usual topology of the complex numbers), and, in my own experience, such a comparison is just as likely to indicate a mistake on the part of the programmer as it is a deliberate design choice. That paper says that orders are deliberately omitted from `std::complex` to avoid inappropriate comparisons. The same is true in Python. On the other hand, NumPy allows comparisons of complex values.

My own inclination thus far is toward eventually providing callables for lexicographic sorts and comparisons (assuming the infrastructure is in-place to do that...), and letting users that want lexicographic sorting of complex numbers get the sorting operations by first viewing the data as pairs of floats. Does that sound reasonable to you?

mwiebe:
@insertinterestingnamehere the problem isn't just complex, though, NaN and -0 vs 0 are themselves problematic as described in that paper. NumPy handles NaN specially within sort, while Python produces incorrect results (-1 stays after 0 when a NaN is between them):

```
In [14]: np.sort([0, np.nan, -np.nan, -1])
Out[14]: array([ -1.,   0.,  nan,  nan])

In [15]: sorted([0, np.nan, -np.nan, -1])
Out[15]: [0, nan, nan, -1]
```

I like what the paper suggests, of having the normal comparison operators be based on the mathematical ideas where possible, and defining new comparison operations `total_less`, `total_equal`, etc, which satisfy total ordering/equivalence relationships so as to be appropriate for sorting, tree-based structures, hash tables, etc.

mwiebe:
Actually, doing a tri-state total_order function like `template<typename T> total_ordering total_order(T,T)` from the paper is more generally useful, especially for composed total orderings like lexicographic comparison of tuples. If we only have one additional ordering function for sort() and other similar algorithms to use by default, this should be it.

izaid:
I'm just +1 to having this discussion, not to any particular side.

insertinterestingnamehere:
@mwiebe I like the system outlined in that proposal. One thing we may still want to consider is providing a IEEE 754-2008 compliant total ordering function for floating point numbers. That paper seems to imply that that is the best way to resolve the issues with floating point numbers, though it doesn't explicitly say so. That would work fine independent of a lexicographic ordering callable. What I had in mind with that was to create something that would perform a lexicographic sort along a specified axis or list of axes by lexicographically comparing slices along the remaining axes. Doing a lexicographic comparison like that for higher dimensional arrays would sometimes require that a path for flat iteration be specified, but such a function would be very useful.

insertinterestingnamehere:
So, summarizing what we've settled thus far (as I understand it):

- Ordering complex numbers is a functionality we do want to provide in some form, possible via a total ordering callable, a lexicographic ordering callable, or both.
- The comparison operators should be disabled by default for complex numbers.

Do we want a type classification corresponding to types that have ordering operators enabled, or should a join be sufficient there?

insertinterestingnamehere:
Okay, I've ammended this to remove the "ordered" list. I'll just join lists to get the needed types for the comparison operators. Other than that, this pretty much follows the C++ naming conventions. We haven't resolved everything to do in the long run, but the discussion here is enough to say what needs to be done right now. We should merge this once the tests pass.

PR #593, opened 2015-10-12
Remove unneeded apt package source.

insertinterestingnamehere:
I noticed a warning about this from the Travis build system in our builds.
The apt source for LLVM 3.4 isn't included in the Travis-ci whitelist for
package repositories, so it isn't used for our builds.


izaid:
Thanks!

PR #601, opened 2015-10-16
Min and max

izaid:


src/dynd/func/min.cpp
```diff
@@ -0,0 +1,34 @@
...
+  auto children = callable::make_all<min_kernel, arithmetic_type_ids>();
+
+  return functional::reduction(
+      functional::multidispatch(ndt::callable_type::make(ndt::scalar_kind_type::make(), ndt::scalar_kind_type::make()),
+                                [children](const ndt::type & DYND_UNUSED(dst_tp), intptr_t DYND_UNUSED(nsrc),
+                                           const ndt::type * src_tp) mutable->callable &
+  {
+                                  callable &child = children[src_tp[0].get_type_id()];
+                                  if (child.is_null()) {
+                                    throw runtime_error("no suitable child found for nd::sum");
```
mwiebe: Error message here says "nd::sum" instead of "nd::min".

PR #602, opened 2015-10-17
Unary operators

insertinterestingnamehere:
This takes care of some of the lower hanging fruit with the operator code. It enables the logical and bitwise not operators. It also contains some initial work toward enabling the other logical operators. Tests to follow.


include/dynd/bool1.hpp
```diff
@@ -40,6 +40,22 @@ class DYND_API bool1 {
...
+
+  DYND_CUDA_HOST_DEVICE bool1 operator&&(bool1 &rhs) {
+    return bool1(m_value && rhs.m_value);
+  }
+
+  DYND_CUDA_HOST_DEVICE bool1 operator||(bool1 &rhs) {
+    return bool1(m_value || rhs.m_value);
+  }
+
   DYND_CUDA_HOST_DEVICE bool1 &operator+=(bool1 rhs)
```
izaid: So, a method in a class is `inline` by default. You actually don't need to specify it here -- if it's specified anywhere else in DyND, it should be removed.

include/dynd/int128.hpp
```diff
@@ -81,7 +81,15 @@ class DYND_API int128 {
...
-  
+
+  DYND_CUDA_HOST_DEVICE inline bool operator!() const {
+    return !(this->m_hi) && !(this->m_lo);
+  }
+
+  DYND_CUDA_HOST_DEVICE inline int128 operator~() const {
+    return int128(~m_hi, ~m_lo);
+  }
+
```
izaid: Same `inline` comment.

include/dynd/uint128.hpp
```diff
@@ -109,6 +109,14 @@ class DYND_API uint128 {
...
+  DYND_CUDA_HOST_DEVICE inline bool operator!() const {
+    return (m_hi != 0) && (m_lo != 0);
+  }
+
+  DYND_CUDA_HOST_DEVICE inline uint128 operator~() const {
+    return uint128(~m_hi, ~m_lo);
+  }
+
   DYND_CUDA_HOST_DEVICE inline bool operator!=(const uint128 &rhs) const
   {
```
izaid: Same `inline` comment.

tests/func/test_arithmetic.cpp
```diff
@@ -282,40 +282,40 @@ TEST(Arithmetic, OptionArithmeticInt32)
...
+  EXPECT_ALL_FALSE(nd::is_avail(1 + NA));
+  EXPECT_ALL_FALSE(nd::is_avail(1 - NA));
+  EXPECT_ALL_FALSE(nd::is_avail(1 * NA));
+  EXPECT_ALL_FALSE(nd::is_avail(1 / NA));
+
+  EXPECT_ALL_FALSE(nd::is_avail(NA + NA));
+  EXPECT_ALL_FALSE(nd::is_avail(NA - NA));
+  EXPECT_ALL_FALSE(nd::is_avail(NA * NA));
+  EXPECT_ALL_FALSE(nd::is_avail(NA / NA));
 }
```
izaid: Why the change to EXPECT_ALL_FALSE? Why doesn't EXPECT_FALSE work anymore?

include/dynd/bool1.hpp
insertinterestingnamehere: Okay, I've seen that all over the place. I'll remove it here, but a style PR is probably in order for everything else.

tests/func/test_arithmetic.cpp
insertinterestingnamehere: `EXPECT_FALSE` implicitly performs a static cast to bool by applying the logical not operator. Once a class overloads the logical not operator, the appropriate overload is called instead, so an array is returned from the expression rather than a boolean value. I didn't finish digging through the gtest code to find exactly why it fails after that. I could try fixing it in their code.

izaid:
Yeah, so Appveyor is obviously failing here. There's no rush, but we can't merge this until that is fixed.

insertinterestingnamehere:
Yep. MSVC is warning about an implicit cast from integer to bool. It's very easy to fix. I've been focusing primarily on the design doc for the operators, but this is easy to finish. I'll try to get fixes up later today.

insertinterestingnamehere:
Oddly enough, the error is only present on AppVeyor. My installation of MSVC 2015 has no complaints.

insertinterestingnamehere:
Okay, rebased and fixed outstanding issues. I didn't mess with the testing stuff yet though since it's not really clear what we want to do there. Assuming the tests pass, this should be good to go.

PR #607, opened 2015-10-18
Changed internal representation of variable bytes and strings

izaid:


tests/types/test_bytes_type.cpp
```diff
@@ -17,86 +17,90 @@
...
-    EXPECT_EQ(1u, d.p("target_alignment").as<size_t>());
-    EXPECT_FALSE(d.is_expression());
-    // Roundtripping through a string
-    EXPECT_EQ(d, ndt::type(d.str()));
+  // Strings with various alignments
+  d = ndt::bytes_type::make(1);
+  EXPECT_EQ(bytes_type_id, d.get_type_id());
+  EXPECT_EQ(bytes_kind, d.get_kind());
+  EXPECT_EQ(alignof(bytes), d.get_data_alignment());
+  EXPECT_EQ(sizeof(bytes), d.get_data_size());
```
mwiebe: Might be nice to keep these tests in terms of `sizeof(void *)` or similar, because the intention is to define the specific layout in memory so C code or other JITted could could interoperate without having the C++ definition of `bytes`.

izaid:
I think we should test for this, but the right way is probably to add a `static_assert` that checks `alignof` and `sizeof` for `dynd::bytes` and `dynd::string`.

PR #620, opened 2015-10-26
[WIP] Sort

izaid:
This is a first attempt at using std::sort in DyND.


izaid:
I'm merging this now. It's not totally done, but it's a good start. The further changes we need are somewhat connected to bytes and string refactoring.

Unfortunately, it's not as trivial as I hoped to hook into STL algorithms, but I'm optimistic we can get there.

PR #649, opened 2015-11-04
Partial progress towards new neighborhood

izaid:


izaid:
This is a pretty big refactoring of nd::functional::neighborhood, a very neat functional of ours that has been sorely left behind. It's only a partial refactoring so far, there will be further updates.

PR #657, opened 2015-11-06
Fix various build errors when building with MSVC 2015 and Python 3.

insertinterestingnamehere:
`timezone` is used as a macro in the Python 3.5 headers. Its declaration
there was making it so that, rather than getting arguments with a given
type and the name `timezone` we were getting arguments with an alternative
type formed by parsing the type we specified and the expansion of the
`timezone` macro. The simplest solution for now was to just not use
`timezone` as a name for a function argument. A more robust
preprocessor-based solution can be added later.

### DIFF
--- a/include/dynd/types/datetime_type.hpp
+++ b/include/dynd/types/datetime_type.hpp
@@ -30,7 +30,7 @@ namespace ndt {
     datetime_tz_t m_timezone;
 
   public:
-    datetime_type(datetime_tz_t timezone);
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
See https://github.com/python/cpython/blob/master/PC/pyconfig.h#L230. The corresponding macro in the MSVC headers was `(*__timezone())` so our arguments named `timezone` were being transformed into unnamed function pointers.

PR #679, opened 2015-11-16
Make sure dll in bin directory is installed at install time

insertinterestingnamehere:


insertinterestingnamehere:
@mwiebe This should work now. Any thoughts?

mwiebe:
Building the tests is optional, so having the .dll be installed only if tests are built doesn't quite seem right.

insertinterestingnamehere:
Good point. I've found a better solution now. Thanks.

PR #690, opened 2015-11-18
Document the cuda option in the libdynd-config batch script.

insertinterestingnamehere:


insertinterestingnamehere:
Yes, the main reason I'm looking at this now is that the handling of configuration options from libdynd needs to be cleaned up a bit to get the conda build for dynd-python working.

PR #696, opened 2015-11-20
Added a replacement for gfunc property access

izaid:


include/dynd/memblock/array_memory_block.hpp
```diff
@@ -38,7 +38,7 @@ struct DYND_API array_preamble : memory_block_data {
       if (!owner) {
         // Call the data destructor if necessary (i.e. the nd::array owns
         // the data memory, and the type has a data destructor)
-        if ((tp->get_flags() & type_flag_destructor) != 0) {
+        if (!tp->is_expression() && (tp->get_flags() & type_flag_destructor) != 0) {
```
mwiebe: This change looks a bit fishy to me, why was it necessary? An expression type's storage type could require destruction, and it should be inheriting this particular flag from the storage type, not the value type.

izaid:
I found that the property type was having it's data_destruct. I'd like to walk some of these changes back, but I'd like to do that after we remove gfunc once and for all. Extracting gfunc without breaking everything has proven to be pretty challenging.

PR #724, opened 2015-11-30
[WIP] Moving towards a call(array *, array *const *) by default

izaid:


izaid:
Closing this, this is a good idea but requires multiple steps to execute on.

PR #835, opened 2015-12-22
Trying new reference counting strateg

izaid:


izaid:
This was a failed attempt.

PR #871, opened 2016-01-10
callable static initialization fix

insertinterestingnamehere:
This fixes a bug that came to light in dynd-python where the struct type and the corresponding static data for each callable exported using the declfunc interface was getting included by the linker on a per-dll basis rather than being instantiated once in libdynd and shared across modules that depend on it.


insertinterestingnamehere:
@izaid The segfault that's showing up in the Travis tests is already in master. For some reason the CI tests didn't pick up on it until now. I restarted https://travis-ci.org/libdynd/libdynd/jobs/100152287 and it's failing as well even though it originally passed. Any ideas? Other than that, this should be good to go.

izaid:
Awesome work, thanks for sorting this out!

PR #873, opened 2016-01-11
Create a serialization design document as a starting point

mwiebe:
Tried to cover the motivations and design constraints in just enough detail to get started.

### DIFF
--- a/docs/serialization-design.md
+++ b/docs/serialization-design.md
@@ -18,8 +18,6 @@ As a serialization protocol, this document is concerned with converting to/from 
 
 The `serialize` callable has signature `(Any) -> bytes`, and operates as a reduction kernel. This means it concatenates data to the output bytes object. Only the data gets serialized, not the type. This is not a problem, however, because to serialize an arbitrary array, one can serialize it as `array[Any]`, which must serialize both the type and the data.
 
-All primitives are serialized in little-endian order. By making this choice explicitly, there is no ambiguity across different platforms, and by not allowing alternatives, this makes the serialization suitable for use in Merkle DAGs. A majority of platforms are little-endian, so this is also good for the typical case.
-
 In DyND's storage within memory, there is some metadata which describes memory layout, data references, and other details incidental to how things are stored. For serialization, the layout is defined completely in terms of the type, and there is no such metadata. For example, the `fixed` dimension type has a stride in its metadata. In the serialization of a strided array, the data is always tightly packed, and no stride is stored anywhere.
 
 ## Deserialize Callable
@@ -27,6 +25,12 @@ In DyND's storage within memory, there is some metadata which describes memory l
 The `deserialize` callable has signature `(bytes, dst_tp: type) -> Any`, where the output array has the specified destination type. The generic case of of deserializing a general DyND array is treated as an element of type `array[Any]`, as noted in the serialize section above.
 
 ## Type to Bytes Mapping
+
+A fundamental principle used to design this mapping is that the serialization of particular data for a particular type must be unique. This is not the case in DyND's storage within memory, where there is some metadata which additionally describes memory layout and some other details. The reason for this design choice is to allow this serialization to be a building block in CAS/Merkle DAG technologies.
+
+One choice made to remove this ambiguity is that all primitives are serialized in little-endian order. A majority of platforms are little-endian, so this is also good for the typical case.
+
+A consequence of this choice is that multi-dimensional arrays always serialize in "C order". This is great for deduplication of data, but does have some performance consequence when working with "F order" arrays. I (Mark) view the support of the CAS/Merkle use case to be strong enough to not change the unique serialization principle, and propose doing something within the type system layered on top of the simple core serialization definition for cases where the performance is more critical. For example, a type which represents a storage axis permutation could do this in general, like `storage[3 * 5 * 7 * float64, axis=[2, 1, 0]]` or `storage[3 * 5 * 7 * float64, order='F']` could represent F order.
 
 * `bool`
   * Single byte `0x00` (false) `0x01` (true)
@@ -41,7 +45,7 @@ The `deserialize` callable has signature `(bytes, dst_tp: type) -> Any`, where t
 * `string`
   * The UTF-8 encoding of the string is serialized identically to `bytes`.
 * `char`
-  * The 32-bit code point is serialized as is. (TODO: Would be better as UTF-8?)
+  * The code point is serialized as its UTF-8 bytes. This means arrays of `char` serialize into a UTF-8 representation.
 * `type`
   * The type represented as a string, serialized as `string`. (TODO: Need to specify a well-defined type to string serialization that we will never change.)
 * `void`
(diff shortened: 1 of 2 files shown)


mwiebe:
Note that the choice of string/bytes serialization is something I first tried here: http://docs.thinkboxsoftware.com/products/sequoia/1.0/Documentation/manual/formats/prt2.html and is working well. If you read this, you'll see that it has been designed in terms of a subset of the DyND type system with the idea of future expansion.

insertinterestingnamehere:
Reading over this quickly, it looks good. I have two quick thoughts, though they don't really need to be added to the design doc.

1. It's a pity we can't give at least a little information about the memory layout. It would make this a little more complicated, but the ability to preserve contiguity along a given axis would be nice.
2. At some point we should add hooks that allow dynamically registered types to specify how they are serialized. It should be possible to make it so that if a C++ type we're registering provides the right serialization and deserialization interfaces we automatically register it with the needed callables when the new type is added to our registry.

docs/serialization-design.md
```diff
@@ -0,0 +1,64 @@
...
+
+We are not concerned with deduplication, as that would complicate the serialization format a fair bit. Deduplication can be handled by a future extension to a new type like `array[Any]`, but which operates as a Merkle link, together with serialization to/from a content addressable storage (CAS).
+
+We are not concerned with compression, that is something that can be layered on top of serialization. If a stream interface is added to DyND, chaining compression to serialization via that interface would be natural to pursue.
+
+As a serialization protocol, this document is concerned with converting to/from byte streams, and  not with random access or efficient seeking. In some cases, random access is possible, however.
+
+## Serialize Callable
+
+The `serialize` callable has signature `(Any) -> bytes`, and operates as a reduction kernel. This means it concatenates data to the output bytes object. Only the data gets serialized, not the type. This is not a problem, however, because to serialize an arbitrary array, one can serialize it as `array[Any]`, which must serialize both the type and the data.
```
mrocklin: The phrase "only the data gets serialized, not the type" concerned me.  It looks like you follow this with a reason about why this isn't a problem.  Just want to verify that the serialization would be self-describing.

docs/serialization-design.md
```diff
@@ -0,0 +1,64 @@
...
+* `var * T`
+  * The size is stored as a [base 128 Varint](https://developers.google.com/protocol-buffers/docs/encoding?hl=en#varints), and then the elements are serialized the same as `N * T`.
+* `(T0, T1, ...)`, `{field0: T0, field1: T1, ...}`
+  * Each element of the tuple/struct is serialized in sequence (a packed contiguous serialization).
+* `?T`
+  * For types where the missing data sentinel is well-defined, the data is serialized just as `T` would be. For other types, missing is serialized as a single byte `0x00`, and available data is serialized as a single byte `0x01` followed by the serialization as `T`.
+* `pointer[T]`
+  * Serialized exactly as `T` would be.
+* `array[T]`
+  * Serialized as if it were a tuple `(type, T)`, with the type serialized, immediately followed by the data.
```
mrocklin: As people add custom types they'll also want a mechanism to serialize and deserialize those types.  What is a good solution here?

mrocklin: Thoughts on categoricals?

docs/serialization-design.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Possible DyND Array Serialization As CBOR
+
+The serialization could be as [CBOR](http://cbor.io/), through the mechanism described in http://tools.ietf.org/html/rfc7049#section-7.2. This is not thought out properly yet.
```
mrocklin: Are current CBOR standards sufficient to describe all dynd arrays?  I know it's community-extensible, I just don't know how far along they've gotten in defining their standards.

docs/serialization-design.md
mrocklin: Ah, I see that @insertinterestingnamehere already made the point about custom types.

izaid:
Okay, just picking off the custom type comment because both @insertinterestingnamehere and @mrocklin. It's already handled in general. All of DyND's callables are, by default, overloadable for user-defined types. One can create a user-defined type dynamically via the type registry and add overloads as appropriate. This set of features already exists and works independently within DyND.

For serialization, one would simply write a serialization callable with the appropriate signature `(SomeType) -> bytes` and do `nd::serialize.set_overload(...)`.

mwiebe:
@insertinterestingnamehere for the point 1 about memory layout, @izaid and I have talked about using types like https://github.com/libdynd/libdynd/blob/master/include/dynd/types/type_id.hpp#L150 to enforce this kind of thing. Does this feel like it gets at what you're imagining?

docs/serialization-design.md
mwiebe: Another reason, which I didn't write down, is that it has to be this way to be able to define the serialization properly. The type generally specifies the structure of the data quite rigidly, unlike JSON-like serialization formats where each element can have different types. If we put the type into the serialization of everything, we can't recursively apply that definition for serializing arrays, tuples, etc, without the type appearing repeatedly everywhere.

docs/serialization-design.md
mwiebe: CBOR is primarily designed to serialize JSON data. As such, there is one "number" type, among other things. Whether we can interact well with that data model is a thing that is not thought out, this may turn out to be a bad idea.

In http://cbor.io/, it says "CBOR is defined in an Internet Standards Document, RFC 7049. The format has been designed to be stable for decades."

insertinterestingnamehere:
@mwiebe it feels like a start, though it'd be nice to be able to specify which axis is contiguous rather than just requiring that it be the first or the last one. Ideally there would be a way to mark any particular dimension as contiguous rather than making that be enforced by some syntax layered onto what is, conceptually, the item type. That issue definitely shouldn't hold up our developing a serialization interface though.

docs/serialization-design.md
seibert: Looking through the CBOR docs, it appears they follow the JSON data model structurally, but also define the usual set of specialized binary integer and float types one would expect.

mrocklin:
If we don't plan to access the bytes then what is the advantage of specifying axis linearization order?  
Is this for serialization/deserialization efficiency?

There is a cost of ambiguity to adding extra knobs.  From a deduplication perspective we want any two dynd arrays with equal values to have equal serializations.

docs/serialization-design.md
mwiebe: They have a subsection http://tools.ietf.org/html/rfc7049#section-3.6 which delves a bit more into recommendations about interpreting this.

seibert: I see, so DyND would basically specify a CBOR protocol where different numeric data formats would be used based on the datashape.

insertinterestingnamehere:
@mrocklin the main benefit I see here is that users would get a memory layout back from the serialized data that roughly matches the one they put in. It would make it so that people whose code is designed to work with a given memory layout wouldn't have to restructure their data in memory every time it is deserialized. That's a choice in both the efficiency and in the interface. That said, it's not a huge issue, just something I think would be nice to have. Distinctions like this are more important in working with our callables than they are for serialization.

mrocklin:
I can see value in that.  In many cases deduplication is not as significant a concern as is serialization/deserialization bandwidth.  I think that decisions here should be informed by how slow serialization along a mis-aligned dimension is compared to I/O channels.

mwiebe:
Using NumPy to give an example of what kind of performance differences you might expect:

```
In [24]: x = np.empty(10000000).reshape(100, 100, 100, 10)

In [25]: timeit x.copy(order='C')
10 loops, best of 3: 30.3 ms per loop

In [26]: timeit x.copy(order='F')
10 loops, best of 3: 140 ms per loop

In [27]: timeit np.save('x.npy', x)
1 loops, best of 3: 694 ms per loop
```

mwiebe:
I tweaked this document a bit in a new commit. Having thought a bit more about preservation of memory order, I really like having the uniqueness property at the base serialization level, and believe it's necessary for DyND's serialization to work well as a primitive in building Merkle DAG systems.

PR #883, opened 2016-01-24
Create merkledag-design.md, doc for Merkle DAG stuff

mwiebe:
This is a first revision of some ideas about how DyND can play a role as infrastructure for building Merkle DAG systems. It attempts to explain some common ways such systems can be used, and includes a first stab at some of the specifics of how the storage of nodes and data in a CAS might look.


izaid:
Merging this in.

PR #887, opened 2016-01-24
Use "-O0 -g" in debug build (cmake -DCMAKE_BUILD_TYPE=debug).

skrah:
Hi,

would it be possible to use "-O0 -g" in the debug build?  Cmake overrides CFLAGS="-O0 -g"
passed on the command line, so changing CMakeLists.txt seems to be the only sane way.


izaid:
Yep, will merge ASAP. Thanks!

PR #898, opened 2016-01-28
Throw an informative error when integer division or mod by zero occurs.

insertinterestingnamehere:


include/dynd/kernels/arithmetic.hpp
```diff
@@ -68,7 +68,49 @@ namespace nd {
...
+  namespace detail {                                                                                                   \
+    template <type_id_t Src0TypeID, type_id_t Src1TypeID,                                                              \
+              type_id_t Src0Base = base_id_of<Src0TypeID>::value,                                                      \
+              type_id_t Src1Base = base_id_of<Src1TypeID>::value,                                                      \
+              bool check = ((Src0Base == bool_kind_id) || (Src0Base == int_kind_id) || (Src0Base == uint_kind_id)) &&  \
+                           ((Src1Base == bool_kind_id) || (Src1Base == int_kind_id) || (Src1Base == uint_kind_id))>    \
+    struct inline_##NAME;                                                                                              \
+                                                                                                                       \
+    template <type_id_t Src0TypeID, type_id_t Src1TypeID, type_id_t Src0Base, type_id_t Src1Base>                      \
```
izaid: Looks good, this is the right way to do it. For consistency, I've been doing Src0ID, Src0BaseID, Src1ID, Src1BaseID rather then Src0ID, Src1ID, Src0BaseID, Src1BaseID. We should probably stick with the former as it is used throughout DyND.

include/dynd/kernels/arithmetic.hpp
```diff
@@ -68,7 +68,49 @@ namespace nd {
...
+      static R f(T0 a, T1 b) {                                                                                         \
+        if(b == 0) {                                                                                                   \
+          throw dynd::zero_division_error("Integer division or modulo by zero.");                                      \
+        }                                                                                                              \
+        return a OP b;                                                                                                 \
+      }                                                                                                                \
+    };                                                                                                                 \
+                                                                                                                       \
+    template <type_id_t Src0TypeID, type_id_t Src1TypeID, type_id_t Src0Base, type_id_t Src1Base>                      \
```
izaid: Rather then the check template variable, what I would do here is break out the int_kind and bool_kind into a template specialization. This might require some demacroizing....

include/dynd/kernels/arithmetic.hpp
insertinterestingnamehere: I did things in that order to avoid adding an extra layer of indirection in the templates before passing the kernel in to `make_all` at https://github.com/libdynd/libdynd/blob/master/include/dynd/func/arithmetic.hpp#L61. Is there another interface there that I should be using or are you asking for another set of templates here?

izaid: The solution we've been using here is a template alias -- see https://github.com/libdynd/libdynd/blob/master/include/dynd/kernels/uniform_kernel.hpp#L229.

I'm also open to other solutions, anything that simplifies.

include/dynd/kernels/arithmetic.hpp
insertinterestingnamehere: I did that deliberately since the code is completely identical for unsigned integers, signed integers, and bools and we have to check that both arguments are one of those types. That's why I originally implemented this with `std::is_integral`. Is there a another way to do this that doesn't involve rewriting the same thing nine times?

include/dynd/kernels/arithmetic.hpp
insertinterestingnamehere: An alias like that looks good to me. I'll add that in. Thanks.

include/dynd/kernels/arithmetic.hpp
izaid: If we leave the type ID as an unspecialized template variable, but specialize on the base ID for unsigned, signed, and bool, then we only have to write it out 3 times. I think you can probably simplify, but you just need to play around a bit. This is partially a question of personal preference, I think.

insertinterestingnamehere:
Okay, here's a newer version that I like a lot better. Rather than keep the handling of the different cases for bool vs int vs unsigned int in the macro itself I pulled it out as a constexpr function. This made it so that the extra template parameters for the base types weren't necessary and also made it so that the logic on when to check for zeros could be put in a function body instead of a template signature.

izaid:
I'm happy with this, merging, thanks!

insertinterestingnamehere:
Great! Thanks for the review!

PR #901, opened 2016-01-28
WIP: First pass at simple string concatentation kernel

mdboom:
This isn't a pull request, so much as a way to communicate my questions as I learn...

This implements the string concatenation kernel for starters.  At the moment it's not templatized, but works *only* on `string_id`, not the `fixed_id` or `char_id` types.  My understanding is that's sort of the implementation plan (since even something simple like concat would require extra checking and different semantics for fixed-length strings).  If it makes sense for consistency to templatize it anyway (with only one implemented specialization) I can -- just trying to keep it as simple as possible for now.

It looks like the existing string concat kernel took *N* input arguments.  I'm not quite clear how to implement that based on what I'm seeing in other kernels.  (Am I right that the last template argument to `base_kernel` is the number of inputs to the kernel, so it's kind of compile-time hardcoded...?)  If variable number of arguments is important to have, is there another kernel whose example I could follow?

For convenient use, I assume there will need to be some elwise functions predefined somewhere...  At the moment, I just threw this into the testing function:

```
    nd::callable f = nd::functional::elwise(nd::callable::make<nd::string_concatenation_kernel>());
```

Is there a convention/best practice for that?  I looked at the instantiation for `assign_na` in `option.cpp`, but was left a little baffled...

Do we want to retain the `operator+` syntax for string concatenation in the existing test?  (I know that convention is very popular among programming languages, but given how different it is from actual addition, I don't know if that's an example we *have* to follow).


include/dynd/kernels/string_algorithm_kernels.hpp
```diff
@@ -3,83 +3,33 @@
...
-    ckernel_prefix m_base;
-    // The number of input operands
-    size_t m_nop;
+      void single(char *dst, char *const *src);
+    };
 
-    ckernel_prefix& base() {
-        return m_base;
-    }
+  } // namespace nd
```
izaid: Your string_concatenation kernel is perfect. You just need to fill in the body of single with the operation, and that's exactly how we write them.

include/dynd/kernels/string_algorithm_kernels.hpp
```diff
@@ -3,83 +3,33 @@
...
-    static void single(char *dst, char *const *src,
-                       ckernel_prefix *extra);
-    static void strided(char *dst, intptr_t dst_stride, char *const *src,
-                        const intptr_t *src_stride, size_t count,
-                        ckernel_prefix *extra);
-};
+    template<>
+    struct traits<dynd::nd::string_concatenation_kernel> {
+      static type equivalent() { return callable_type::make(make_type<dynd::string>(), {type(string_id), type(string_id)}); }
+    };
```
izaid: The `ndt::traits<...>` template that you are using communicates the signature of the kernel to how we make callables (details to follow on that).

tests/types/test_string_type.cpp
```diff
@@ -427,30 +429,29 @@ TEST(StringType, Comparisons)
...
 
-/*
-TODO: Reenable this.
 
 TEST(StringType, Concatenation) {
     nd::array a, b, c;
 
+    nd::callable f = nd::functional::elwise(nd::callable::make<nd::string_concatenation_kernel>());
+
     a = "first";
```
izaid: How you made the `nd::callable f` is also good. There's a way we declare these functions in DyND as global variables, which is all you are missing.

izaid:
Hey @mdboom! I'm impressed with your very quick start. So far, so good. Here are my answers to your questions.

Yes, we are only implementing those operations on variable-length strings (the default `dynd::string`), so there is no need to template these kernels on type IDs. I have no idea what the existing string concat kernels were doing, they have been disabled for a long time. Just delete them all and ignore what they were.

You've made the callable correctly, so that's great. I know things like `assign_na` in option.cpp are confusing, I think that comes from having to explode the kernels for all these different types. In this case, we don't need to do that because we are just operating on strings. So one very simple example of how to make a DyND global callable would be https://github.com/libdynd/libdynd/blob/master/include/dynd/func/view.hpp and https://github.com/libdynd/libdynd/blob/master/src/dynd/func/view.cpp. You basically just need to declare a `string_concatenation` struct with a `make` function that literally returns exactly `nd::functional::elwise(nd::callable::make<nd::string_concatenation_kernel>())`. This function can then be called like a normal DyND callable.

In terms of code structure, what we like to do is implement one kernel (or one kernel template) per file in the kernels directory. We then aggregrate those into files in the main `dynd` directory where those global callables get created. So, in this case, we'd put `string_concat_kernel.hpp` in `dynd/kernels` and then we'd have a `dynd/string.hpp` that declares operations like `string_concat`, `substring_find`. In `dynd/string.cpp`, we include all the kernel files and define all the make functions.

Does all of that make sense?

mdboom:
Thanks for the feedback.  I've reorganized according to your comments.

To confirm: it's ok that `string_concatenation` always takes only 2 operands for now?  (If the framework supports `N` operands, this would be an obvious candidate for that).

Tried to get `a + b` working on arrays of strings by adding `string_id` to the set of binop types passed here: https://github.com/libdynd/libdynd/blob/master/include/dynd/func/arithmetic.hpp#L117 but that didn't work...  I have everything compiling, but I get `no viable overload for dynd::nd::add with argument types "string" and "string"` when running the test.

izaid:
Re: 2 operands versus N operands. It probably is a good idea to support N operands, but I'd leave it at 2 for now solely because it's very slightly awkward with our dispatch system. You could actually write the kernel for that without a problem, it's just the dispatch mechanics.

include/dynd/string.hpp
```diff
@@ -0,0 +1,19 @@
+//
+// Copyright (C) 2011-15 DyND Developers
+// BSD 2-Clause License, see LICENSE.txt
+//
+
+#pragma once
+
+#include <dynd/callable.hpp>
+#include <dynd/kernels/string_concat_kernel.hpp>
+
```
izaid: `#include <dynd/kernels/string_concat_kernel.hpp>` should only be in the *.cpp file. That really simplifies the amount of includes visible to the user.

mdboom: Makes sense.

mdboom:
I've removed the "WIP" tag here.

izaid:
Fantastic job, thanks a lot! Merging.

PR #903, opened 2016-01-29
Make tuple_type.m_field_types a vector (hopefully this can be condensed later).

skrah:
I don't consider this production ready, but it passes the tests and is Valgrind-clean.

As I understood, a pull request makes it easier for you to discuss changes, so any
comments on how to shorten the patch are very welcome!


include/dynd/callable.hpp
```diff
@@ -155,9 +155,9 @@ namespace nd {
 
     std::intptr_t get_narg() const { return get_type()->get_npos(); }
 
-    const ndt::type &get_arg_type(std::intptr_t i) const { return get_type()->get_pos_type(i); }
+    const ndt::type get_arg_type(std::intptr_t i) const { return get_type()->get_pos_type(i); }
 
-    const array &get_arg_types() const { return get_type()->get_pos_types(); }
+    nd::array get_arg_types() const { return get_type()->get_pos_types(); }
```
insertinterestingnamehere: There are a variety of places here where you change const references to by value returns. Is there any particular reason for this?

include/dynd/kernels/get_then_copy_kernel.hpp
```diff
@@ -11,7 +11,40 @@
 namespace dynd {
 namespace nd {
 
-  template <typename TypeType, const array &(TypeType::*Func)() const>
+template <typename TypeType, array (TypeType::*Func)() const>
+struct get_then_copy_kernel2 : base_kernel<get_then_copy_kernel2<TypeType, Func>, 0> {
```
insertinterestingnamehere: Does this completely replace the old `get_then_copy_kernel`? Why not make the changes to the original rather than having two kernels?

skrah: Yes, the goal is to keep one version only.  Currently both versions are needed since other parts still need to be converted.

include/dynd/callable.hpp
skrah: It would return a reference to a local object otherwise:

/home/stefan/libdynd/include/dynd/callable.hpp:160:53: error: returning reference to local temporary object [-Werror,-Wreturn-stack-address]
    const nd::array &get_arg_types() const { return get_type()->get_pos_types(); }

insertinterestingnamehere: Yes, that error makes sense since `get_pos_types` doesn't return a const reference either (nor does `get_field_types`). The change would have to be made for all of them.

insertinterestingnamehere:
I'm still reading through this since it's a part of our code that I haven't looked through before, but thus far this looks like a great improvement. Thanks for taking care of this!

insertinterestingnamehere:
Also, quick recommendation, you may want to build stuff like this on a feature branch rather than on your master branch. That helps you keep a clean copy you can branch off of if you need to work on something else.

include/dynd/callable.hpp
mwiebe: We might want to change these to return a const reference to the internal std::vector instead? Implicitly doing a copy of the whole array when get_arg_types() and other API is called feels like it adds hidden performance traps.

include/dynd/types/tuple_type.hpp
```diff
@@ -145,13 +147,13 @@ namespace ndt {
     }
 
     /** Makes a tuple type with the specified types */
-    static type make(const nd::array &field_types, bool variadic = false)
+    static type make(std::vector<type> field_types, bool variadic = false)
     {
       return type(new tuple_type(field_types, variadic), false);
```
mwiebe: Should maybe be `new tuple_type(std::move(field_types), ...)`, to avoid the extra copy.

src/dynd/types/tuple_type.cpp
```diff
@@ -15,11 +15,11 @@
 using namespace std;
 using namespace dynd;
 
-ndt::tuple_type::tuple_type(type_id_t type_id, const nd::array &field_types, flags_type flags, bool layout_in_arrmeta,
+ndt::tuple_type::tuple_type(type_id_t type_id, std::vector<type> field_types, flags_type flags, bool layout_in_arrmeta,
                             bool variadic)
     : base_type(type_id, variadic ? kind_kind : tuple_kind, 0, 1,
                 flags | type_flag_indexable | (variadic ? type_flag_symbolic : 0), 0, 0, 0),
-      m_field_count(field_types.get_dim_size()), m_field_types(field_types),
+      m_field_count(field_types.size()), m_field_types(field_types),
```
mwiebe: Probably should be `m_field_types(std::move(field_types))`

src/dynd/types/tuple_type.cpp
```diff
@@ -58,7 +58,7 @@ ndt::tuple_type::tuple_type(type_id_t type_id, const nd::array &field_types, fla
   m_arrmeta_offsets.flag_as_immutable();
 }
 
-ndt::tuple_type::tuple_type(const nd::array &field_types, bool variadic)
+ndt::tuple_type::tuple_type(std::vector<type> field_types, bool variadic)
     : tuple_type(tuple_id, field_types, type_flag_none, true, variadic)
```
mwiebe: `std::move(field_types)` here too, and also anywhere else the by-value std::vector<type> parameter is.

include/dynd/callable.hpp
skrah: Yes, I'll try.  In general there are still many conversions from std::vector -> nd::array because all the consumers expect an nd::array.  Can these be eliminated in the end?

mwiebe: Yes, I think those will generally go away, and consumers changed to expect `std::vector<T>`.

skrah: Okay.  Regarding the const reference to the internal std::vector:  I'm not sure yet what the ownership invariants are.  If everything is a reference to the internal vector of the tuple type, what prevents the tuple from disappearing while something still holds a reference (I'm is usually programming in C, this might be an elementary thing in C++).

Is it an option to use only the move semantics you mentioned below, i.e. not use references at all?

mwiebe: Nothing prevents the tuple from disappearing, that's true. It's the responsibility of the code getting that property to make sure the lifetime of that reference is not longer than the reference of the object it came from.

This is pretty common - e.g. in std::vector, operator[] returns a `T&` reference with exactly the same considerations.

skrah:
Thanks, a feature branch is probably a very good idea.

include/dynd/callable.hpp
mwiebe: Re: the move semantics, for property access using const references is better, as that way the data doesn't have to be copied. To use move semantics, a copy of that internal data would be what is moved, and it's not good to have such an expensive operation in there for just property access.

include/dynd/types/tuple_type.hpp
skrah: Is it worth changing the 'make' signature to take an rvalue reference?

/** Makes a tuple type with the specified types */
static type make(std::vector<type> &&field_types, bool variadic = false)
{
  return type(new tuple_type(std::move(field_types), variadic), false);
}

The drawback is that this would require quite a bit of code changes, like https://github.com/libdynd/libdynd/blob/master/src/dynd/types/datashape_parser.cpp#L953 ::

  ndt::type pos = ndt::tuple_type::make(std::move(pos_arg_types));
  ndt::type kwds = ndt::struct_type::make(kw_names, kw_arg_types);
  ndt::type result_tp = ndt::tuple_type::make(std::move({pos, kwds}));

izaid: Eventually, yes, but I wouldn't worry about these kind of optimizations for now. Let's get our code structure into a good place, then we can optimize that kind of thing.

skrah: Are you also okay with leaving everything a const std::vector<type> for now and
do all the std::moves() later?  -- I've experimented with some small programs
and clang/gcc produced exactly the same asm for move semantics and const-only
vectors (it may be different of course for the actual code here).

izaid: Yes, I'm okay with that. I do think we should do const std::vector & now, because that makes a lot of things easier (with regard to passing temporaries and the like).

skrah: Ah yes, for returning vectors from e.g. get_field_types() I have const std::vector & everywhere now, for constructing tuples I have const std::vector.

izaid: If I understand you correctly, constructing tuples can also be `const std::vector &`, rather then `const std::vector`. It just can't be stored as a reference `&. So the constructor can take a ` const std::vector &`, but store it as a `std::vector` or `const std::vector`.

skrah: You are right (I was still thinking in C).

skrah:
As Ian suggested, I'll switch to a feature branch now, so I guess I'll have to open a new PR soon (see: https://github.com/isaacs/github/issues/18).

PR #905, opened 2016-01-29
Add tests for scalar string concatenation

mdboom:
This probably should have been part of #901.

Just adds a test for the `operator+` I added to `bytes`, and also adds an `operator+=`.

### DIFF
--- a/include/dynd/bytes.hpp
+++ b/include/dynd/bytes.hpp
@@ -89,14 +89,22 @@ public:
 
   const bytes operator+(const bytes& rhs)
   {
-      bytes result;
+    bytes result;
 
-      result.resize(size() + rhs.size());
+    result.resize(size() + rhs.size());
 
-      DYND_MEMCPY(result.begin(), begin(), size());
-      DYND_MEMCPY(result.begin() + size(), rhs.begin(), rhs.size());
+    DYND_MEMCPY(result.begin(), begin(), size());
+    DYND_MEMCPY(result.begin() + size(), rhs.begin(), rhs.size());
 
-      return result;
+    return result;
+  }
+
+  bytes &operator+=(const bytes& rhs)
+  {
+    size_t orig_size = size();
+    resize(size() + rhs.size());
+    DYND_MEMCPY(begin() + orig_size, rhs.begin(), rhs.size());
+    return *this;
   }
 };
 
--- a/include/dynd/types/string_type.hpp
+++ b/include/dynd/types/string_type.hpp
@@ -22,6 +22,8 @@ public:
 
   string(const std::string &other) : string(other.data(), other.size()) {}
 
+  string(const bytes &other) : string(other.data(), other.size()) {}
+
   bool operator<(const string &rhs) const
   {
     return std::lexicographical_compare(begin(), end(), rhs.begin(), rhs.end());
--- a/tests/types/test_string_type.cpp
+++ b/tests/types/test_string_type.cpp
@@ -431,6 +431,22 @@ TEST(StringType, Comparisons)
 }
 
 
+TEST(StringType, ConcatenationScalar) {
+    dynd::string a("first");
+    dynd::string b("second");
+
+    dynd::string c(a + b);
+    ASSERT_EQ(dynd::string("firstsecond"), c);
+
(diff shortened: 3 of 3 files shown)


include/dynd/types/string_type.hpp
```diff
@@ -22,6 +22,8 @@ class DYND_API string : public bytes {
 
   string(const std::string &other) : string(other.data(), other.size()) {}
 
+  string(const bytes &other) : string(other.data(), other.size()) {}
+
   bool operator<(const string &rhs) const
```
izaid: Why is `string(const bytes &other)` necessary? It feels strange to me to have a copy constructor in string from it's base class.

tests/types/test_string_type.cpp
```diff
@@ -431,6 +431,22 @@ TEST(StringType, Comparisons)
...
+    dynd::string a("first");
+    dynd::string b("second");
+
+    dynd::string c(a + b);
+    ASSERT_EQ(dynd::string("firstsecond"), c);
+
+    a = dynd::string("foo");
+    a += dynd::string("bar");
+
+    printf("%s\n", a.begin());
```
izaid: Extra printf that we should remove.

izaid:
Looks useful, thanks! Added some comments.

include/dynd/types/string_type.hpp
mdboom: It's to support:

```
dynd::string a("first");
dynd::string b("second");
dynd::string c(a + b);
```

since `operator+` returns bytes.

There might be a better way, of course.

izaid: Ah, okay. There probably is a better way to do it, but I don't know what it is offhand. So let's leave this in and fix it whenever someone feels like playing around with it.

mwiebe: Hey @mdboom thanks for getting into the code, it's awesome to see more faces in the commit logs :)

I think we want to keep a barrier between dynd::bytes and dynd::string, just like in Python 3. That would mean no implicit conversions, and structuring it so that the explicit conversions are very clear.

include/dynd/bytes.hpp
```diff
@@ -89,14 +89,22 @@ class DYND_API bytes {
...
-      return result;
+    return result;
+  }
+
+  bytes &operator+=(const bytes& rhs)
+  {
+    size_t orig_size = size();
+    resize(size() + rhs.size());
+    DYND_MEMCPY(begin() + orig_size, rhs.begin(), rhs.size());
+    return *this;
```
mwiebe: It looks like because the string isn't following the design in https://github.com/libdynd/libdynd/blob/master/docs/string-design.md yet, this can't currently be implemented with an exponential growth strategy for efficient amortized concatenation. What thoughts does everyone have on finishing that up?

izaid: The only thing that's lacking from the design implementation is the inline storage, right? We should definitely finish that at some point, but it doesn't alter how we are building string kernels -- unless I'm missing something.

izaid:
Merging this in for now, with the understanding that `dynd::bytes` and `dynd::string` need some inspection.

include/dynd/bytes.hpp
mwiebe: It looks like the implementation doesn't have a `capacity` in the heap storage mode either. I'll put a little more in that design doc to try to make it more clear.

mdboom: Yeah, I agree there's lots we could do here.  At present, I'm just working within the confines of the current simple implementation.

izaid: @mwiebe To me, the really important thing is that we're building the string kernels off a relatively stable `dynd::string` API. If you think that's not the currrent case, let me know!

mwiebe: It's pretty close, but I think it needs API to distinguish the exponential growth mode, so that can be used appropriately. The ABI is quite a bit different, that's something that needs to be solidified before connecting it with Numba or another JIT. I've made https://github.com/libdynd/libdynd/pull/907 to give one possible way the exponential growth can be expressed, I can turn that into a code PR later if that API looks alright.

PR #907, opened 2016-01-29
Add some more possible implementation details for reserve/resize

mwiebe:


mdboom:
Makes sense.  Once we have an implementation, it would be interesting to see how much performance the short string optimization gives.

PR #910, opened 2016-01-29
WIP: Add search-related string kernels

mdboom:
This adds three string kernels that all share a common "string searching" code: `find`, `replace` and `count`.

The searching algorithm is a rip-off of the one in CPython described [here](http://effbot.org/zone/stringlib.htm).  It uses a 64-bitfield bloom filter to get better-than-linear search characteristics, since it can skip over subsequent characters in many cases.  The reports of "26x" better is probably not true for UTF-8, but in the common ascii case it should be quite good, and shouldn't be worse than naive search at worst.

Some questions:

The result of `find` and `count` is `intptr_t`, but I can't find a type id that matches that.  I used `int64_id` which works by accident on 64-bit machines, but won't be correct on 32-bit.  Is there a type id corresponding to `intptr_t`?

The sentinel value for "not found" in `find` is -1 (just because that's what Python does), but would it be better to use a na value or something?  Maybe not worth the complexity, but thought I'd ask -- no reason to slavishly follow the Python API here.

This needs a reverse find implementation.


include/dynd/kernels/string_count_kernel.hpp
```diff
@@ -0,0 +1,42 @@
...
+    };
+
+  } // namespace nd
+
+  namespace ndt {
+
+    template<>
+    struct traits<dynd::nd::string_count_kernel> {
+      /* TODO: Rather than int64_id, I really want the id that corresponds to
+               intptr_t... */
```
mwiebe: I think what you want is `ndt::make_type<intptr_t>()` instead of `ndt::type(int64_id)`.

include/dynd/kernels/string_concat_kernel.hpp
```diff
@@ -14,27 +15,6 @@
 namespace dynd {
   namespace nd {
```
izaid: Okay, so you probably only need one of {"#include <dynd/string.hpp>", "#include <dynd/types/string_type.hpp>"} and "#include <dynd/kernels/base_kernel.hpp>" here. We normally don't include kernel_builder directly.

include/dynd/kernels/string_concat_kernel.hpp
```diff
@@ -42,7 +22,7 @@ namespace dynd {
         dynd::string *d = reinterpret_cast<dynd::string *>(dst);
         const dynd::string *const *s = reinterpret_cast<const dynd::string *const *>(src);
 
-        detail::concat_one_string(2, d, s);
+        string_concat(2, d, s);
       }
     };
```
izaid: I'd probably just call `string_concat(2, reinterpret_cast<dynd::string *>(dst), reinterpret_cast<string *>(src[0]));` and skip the temporaries, but it's up to you.

include/dynd/kernels/string_count_kernel.hpp
izaid: What Mark said above.

include/dynd/string.hpp
```diff
@@ -8,11 +8,52 @@
...
+
+  /*
+    In string `src`, replace all non-overlapping appearances of
+    `old_str` with `new_str`, storing the result in `dst`.
+  */
+  void string_replace(dynd::string *const dst,
+                      const dynd::string *const src,
+                      const dynd::string *const old_str,
+                      const dynd::string *const new_str);
+
```
izaid: In terms of API design, these functions should probably take references to `dynd::string` or something. It would be nice if these functions made sense and were usable independently of their corresponding kernels -- e.g., so users could apply them in their own DyND functions.

src/dynd/string.cpp
```diff
@@ -5,14 +5,358 @@
...
+        }
+        s = (const char *)candidate;
+        if (self->handle_match(s - haystack)) {
+          return;
+        }
+        s++;
+      }
+    }
+  }
+
```
izaid: `inline` is actually implicit for any methods in a class or struct, so you can drop it.

src/dynd/string.cpp
```diff
@@ -5,14 +5,358 @@
...
+}
+
+DYND_API struct nd::string_find nd::string_find;
+
+DYND_API nd::callable nd::string_replace::make()
+{
+  return nd::functional::elwise(nd::callable::make<nd::string_replace_kernel>());
+}
+
+DYND_API struct nd::string_replace nd::string_replace;
```
izaid: `nd` is also implicit here because these algorithms are in the `nd` namespace

tests/types/test_string_type.cpp
```diff
@@ -455,41 +455,146 @@ TEST(StringType, Concatenation) {
...
+  b = "abc";
+
+  c = nd::string_find(a, b);
+  ASSERT_EQ(ndt::type("4 * intptr"), c.get_type());
+  ASSERT_EQ(4, c.get_shape()[0]);
+  EXPECT_ARRAY_EQ(0l, c(0));
+  EXPECT_ARRAY_EQ(2l, c(1));
+  EXPECT_ARRAY_EQ(-1l, c(2));
+  EXPECT_ARRAY_EQ(-1l, c(3));
 }
```
izaid: You can match a whole array with `EXPECT_ARRAY_EQ` by passing in the values as an `std::initializer_list`. For instance, this should be equivalent to what you have above: `EXPECT_ARRAY_EQ(nd::array{0l, 2l, -1l, -1l}, nd::string_find(a, b))`. The benefit of that is it does everything you have (checks type, shape, and values), but we get it down to a single line.

tests/types/test_string_type.cpp
```diff
@@ -455,41 +455,146 @@ TEST(StringType, Concatenation) {
...
+  c = nd::string_find(a, b);
+  ASSERT_EQ(ndt::type("6 * intptr"), c.get_type());
+  ASSERT_EQ(6, c.get_shape()[0]);
+  EXPECT_ARRAY_EQ(0l, c(0));
+  EXPECT_ARRAY_EQ(1l, c(1));
+  EXPECT_ARRAY_EQ(2l, c(2));
+  EXPECT_ARRAY_EQ(1l, c(3));
+  EXPECT_ARRAY_EQ(-1l, c(4));
+  EXPECT_ARRAY_EQ(-1l, c(5));
+}
```
izaid: Same comment as above.

include/dynd/string.hpp
mdboom: Sure -- they can take references.  Would it make sense to have C-like interfaces as well (i.e. char * and size_t pairs for each string) so they could be used independently of the `dynd::string` type?  Even further, these could be templatized so `std::string` could be used instead (the used part of the API is the same).

include/dynd/kernels/string_concat_kernel.hpp
mdboom: Which `#include` should I include?

tests/types/test_string_type.cpp
mdboom: That's really cool.  Still ramping up on all the neat stuff in C++11...

include/dynd/kernels/string_concat_kernel.hpp
izaid: Fixed above, markup ate the actual filenames.

include/dynd/string.hpp
izaid: I think we should have a simple API, it can be based off of either the C++ API (my preference) or the C API. If we can support both through templates, even better.

include/dynd/kernels/string_concat_kernel.hpp
mdboom: Ah, got it. ;)

include/dynd/string.hpp
mdboom: I think to cover all bases, we can easily support:

- A template that supports *either* a dynd::string or a std::string (or any class that has `begin`, `end`, `size` and `resize`) ... which then calls...
- A C-like API

That avoids the problem of a lot of old-school C++ APIs that generally say "first create an instance of X (which requires some sort of copy), before you can to anything..."

izaid: Sounds good to me!

mdboom: My plan was to create APIs that take `dynd::string` or `std::string` that call lower-level APIs that take `char *`/`size_t` pairs (the C-like interface).

The detail I forgot in that plan is the memory management.  The C APIs will need to allocate and return the result buffer through a `char **` (since the caller won't know what size to allocate).  But then the C++ API will need to create a `dynd::string` out of it, without copying.  We could add an internal API to assign ownership of a pointer to a `dynd::string`, but such a thing doesn't exist in `std::string`, and it probably breaks @mwiebe's plans to add short-string optimization.

So, I think I come down on *just* supporting `dynd::string` and `std::string`, through a template, and not C-like buffers.  But maybe others have better ideas or strong opinions about what we *must* support here.

izaid: Yeah, the current string proposal from @mwiebe originated as an attempt to move away from exactly that. Our old strings were capable of doing things like that, and it got crazy. ;)

So I'm +1 to implementing algorithms on a generic `StringType` template that supports obvious things.

mdboom:
I think I've addressed all the comments so far...

mdboom:
Running the tests in valgrind on Travis is really cool, btw.  Found some problems...

izaid:
Yeah, we need valgrind to keep things working. Very glad it is there.

izaid:
It looks like Travis CI is fine (the DyND Python failure is due to the merging of StefanK's PR last night, which I'm fixing), but Appveyor has issues. If you can fix Appveyor, we should be good to go.

PR #912, opened 2016-02-01
Remove nd::array from types (Part 1).

skrah:
This is the updated version of #903 that should address all comments.

### DIFF
--- a/src/dynd/types/struct_type.cpp
+++ b/src/dynd/types/struct_type.cpp
@@ -17,7 +17,7 @@
 using namespace std;
 using namespace dynd;
 
-ndt::struct_type::struct_type(const nd::array &field_names, const nd::array &field_types, bool variadic)
+ndt::struct_type::struct_type(const nd::array &field_names, const std::vector<type> &field_types, bool variadic)
     : tuple_type(struct_id, field_types, type_flag_none, true, variadic), m_field_names(field_names)
 {
   /*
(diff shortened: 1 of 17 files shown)


src/dynd/types/typevar_type.cpp
```diff
@@ -154,18 +154,16 @@ bool ndt::is_valid_typevar_name(const char *begin, const char *end)
   }
 }
 
-nd::array ndt::make_typevar_range(const char *name, intptr_t count)
+const std::vector<ndt::type> ndt::make_typevar_range(const char *name, intptr_t count)
 {
```
izaid: Should be std::vector<ndt::type>, I don't think const on return makes sense. (Later on, we'll move this to `&&` anyway.)

src/dynd/types/tuple_type.cpp
```diff
@@ -518,8 +516,8 @@ bool ndt::tuple_type::match(const char *arrmeta, const type &candidate_tp, const
 std::map<std::string, nd::callable> ndt::tuple_type::get_dynamic_type_properties() const
 {
   std::map<std::string, nd::callable> properties;
-  properties["field_types"] = nd::callable::make<nd::get_then_copy_kernel<tuple_type, &tuple_type::get_field_types>>(
-      ndt::callable_type::make(m_field_types.get_type(), ndt::tuple_type::make(),
+  properties["field_types"] = nd::callable::make<nd::get_then_copy_kernel2<tuple_type, &tuple_type::get_field_types>>(
+      ndt::callable_type::make(this->get_type(), ndt::tuple_type::make(),
                                ndt::struct_type::make({"self"}, {ndt::make_type<ndt::type_type>()})));
   properties["metadata_offsets"] =
       nd::callable::make<nd::get_then_copy_kernel<tuple_type, &tuple_type::get_arrmeta_offsets>>(
```
izaid: `this->` is unnecessary anyway, you can drop it.

src/dynd/types/substitute_typevars.cpp
```diff
@@ -27,15 +27,16 @@ using namespace dynd;
 /**
  * Substitutes the field types for contiguous array of types
  */
-static nd::array substitute_type_array(const nd::array &type_array, const std::map<std::string, ndt::type> &typevars,
-                                       bool concrete)
+static const std::vector<ndt::type> substitute_type_array(const nd::array &type_array,
+                                                          const std::map<std::string, ndt::type> &typevars,
+                                                          bool concrete)
 {
```
izaid: Let's remove the `const` for now.

src/dynd/types/struct_type.cpp
```diff
@@ -400,11 +402,10 @@ namespace nd {
 
 } // namespace dynd
 
-static nd::array make_self_types()
+static const std::vector<ndt::type> make_self_types()
 {
```
izaid: Let's remove the `const` for now.

izaid:
Thanks @skrah, great work!

PR #913, opened 2016-02-02
get_then_copy_kernel: parameterize over the return type.

skrah:
For further work more return types will be needed (std::vector<uintptr_t> etc.), so
I think it's best to add a return type parameter to the template.

### DIFF
--- a/include/dynd/kernels/get_then_copy_kernel.hpp
+++ b/include/dynd/kernels/get_then_copy_kernel.hpp
@@ -11,42 +11,8 @@
 namespace dynd {
 namespace nd {
 
-  // WIP (remove nd::array from types/*). Both structs are needed during the transition.
-  template <typename TypeType, const std::vector<ndt::type> &(TypeType::*Func)() const>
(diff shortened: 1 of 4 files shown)


izaid:
This looks fine to me. `get_then_copy_kernel` is one of those DyND constructs that was the best I could do at the time, but really should be replaced by something better. So I'm okay with changing it or deleting it or anything that moves it in the right direction.

skrah:
Yes, replacing get_copy_kernel with something simpler would be nice (if possible).  Currently I just need this to avoid a get_copy_kernel3 struct.

izaid:
I think it's definitely possible, and will be easier to do once we remove the dependency of `ndt::tuple_type`, `ndt::struct_type`, and `ndt::callable_type` on `nd::array`. I will add "removing get_then_copy_kernel" to my personal todo list -- and I'm very glad to avoid get_then_copy_kernel3. ;)

izaid:
@skrah Is this something you want merged in now?

skrah:
Yes, that would be great (for the work on tuple_type.m_arrmeta_offsets)!

izaid:
Okay, will do.

skrah:
Thanks!

PR #918, opened 2016-02-03
Remove nd::array from types (Part 2).

skrah:
This makes tuple_type.m_arrmeta_offsets a vector. get_arrmeta_offsets_raw() is no longer used in libdynd and could be removed.

### DIFF
--- a/include/dynd/types/tuple_type.hpp
+++ b/include/dynd/types/tuple_type.hpp
@@ -22,15 +22,14 @@ namespace ndt {
     intptr_t m_field_count;
 
     /**
-     * Immutable contiguous array of field types. Always has type "N * type".
+     * Immutable vector of field types. Always has type "N * type".
      */
     const std::vector<type> m_field_types;
 
     /**
-     * Immutable contiguous array of arrmeta offsets. Always has type "N *
-     * intptr".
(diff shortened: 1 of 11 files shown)


include/dynd/types/tuple_type.hpp
```diff
@@ -22,15 +22,14 @@ namespace ndt {
...
      */
     const std::vector<type> m_field_types;
 
     /**
-     * Immutable contiguous array of arrmeta offsets. Always has type "N *
-     * intptr".
+     * Immutable vector of arrmeta offsets. Always has type "N * uintptr".
      */
-    nd::array m_arrmeta_offsets;
+    std::vector<uintptr_t> m_arrmeta_offsets;
```
izaid: Should we make this `const` as well?

skrah: Yes, definitely.  The showstopper for now is https://github.com/libdynd/libdynd/pull/918/files#diff-d1a413db4b4cdcb4584670f4f43742caL417 (and also the initialization in the constructor should move to a function).

izaid: Makes sense!

izaid:
I'm very happy with this, merging!

PR #923, opened 2016-02-04
Convert m_field_names to vector

skrah:
This time I've removed the unused *_raw() functions, since the signature has
changed  (so it would break the API anyway).

I've created a str_util.hpp header, since there was a duplicate function.


include/dynd/types/str_util.hpp
```diff
@@ -0,0 +1,23 @@
...
+    if (!(('a' <= *c && *c <= 'z') || ('A' <= *c && *c <= 'Z') || *c == '_')) {
+      return false;
+    }
+    for (++c; c != s.end(); ++c) {
+      if (!(('0' <= *c && *c <= '9') || ('a' <= *c && *c <= 'z') || ('A' <= *c && *c <= 'Z') || *c == '_')) {
+        return false;
+      }
+    }
+    return true;
+  }
```
izaid: I don't want this extra file, but I agree that `is_simple_identifier_name` needs to go somewhere else. I'm not sure where...

include/dynd/types/str_util.hpp
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright (C) 2011-15 DyND Developers
+// BSD 2-Clause License, see LICENSE.txt
+//
+
+static inline bool is_simple_identifier_name(const std::string &s)
+{
```
izaid: `static` should be removed here, that should only be used for functions in a `.cpp` file

src/dynd/types/struct_type.cpp
```diff
@@ -48,45 +50,17 @@ ndt::struct_type::~struct_type() {}
...
+
+  for (intptr_t i = 0; i < m_field_count; ++i) {
+    const std::string &name = m_field_names[i];
+    if (name.compare(0, std::string::npos, field_name_begin, size) == 0) {
+      return i;
     }
   }
 
   return -1;
 }
```
izaid: We should just use `std::find(m_field_names.begin(), m_field_names.end(), field_name_begin)` here if we can make that work.

include/dynd/types/str_util.hpp
skrah: There are a couple of string related functions in src/dynd/json_formatter.cpp that could also go into some utility file. It's hard to find a good location.

izaid: Yeah, I agree. We can leave it here for now, just voicing my thoughts.

include/dynd/types/str_util.hpp
izaid: Also, this needs to be in a namespace `dynd`.

include/dynd/types/str_util.hpp
skrah: But are you okay with the extra file for now? Then I can update the patch.

izaid: Yes, we can certainly keep it for now. I don't have a better idea. :)

src/dynd/types/struct_type.cpp
skrah: I'm not sure: ```const char *field_name_begin``` is not guaranteed to be NUL-terminated (I think!), so the size needs to be passed to the comparison.  With a lambda it would be possible, but it would probably not add clarity.

izaid: It's not guaranteed, but it should be NULL-terminated. Otherwise it's kind of a weird thing to do in C++.

We definitely don't need field_name_end though, so just turning that into a size would be good. (I know this is not code you wrote, but we may as well correct it now.)

skrah: Ah, so it's okay to change the function signature? I'll try that.

izaid: Absolutely! You should feel free to do that in general (and I'll feel free to tell you if the particular change is a bad one).

skrah: In the latest patch I just used ```std::string```.  It would be nice to have ```std::buffer``` for temporarily wrapping a ```const char *``` without making a copy, but it does not seem to exist.

izaid:
Looks good to me, thanks! Will merge soon.

src/dynd/types/struct_type.cpp
mwiebe: http://en.cppreference.com/w/cpp/header/experimental/string_view is what you're thinking of, I think it's making its way into C++17

skrah: Thanks! That's what I was looking for.

PR #941, opened 2016-02-15
Add complete field access kernel with tests.

skrah:


skrah:
This should be a working field access kernel.

tests/func/test_struct.cpp
```diff
@@ -15,4 +15,28 @@
 using namespace std;
 using namespace dynd;
 
-// TEST(Struct, FieldAccess) { nd::field_access(nd::as_struct({{"x", 7}, {"y", 0.5}})); }
+
+TEST(Struct, FieldAccess)
+{
+  nd::array s1 = nd::as_struct({{"x", 7}, {"y", 0.5}});
+  EXPECT_EQ(s1, s1);
+
```
izaid: `EXPECT_EQ(s1, s1)` seems like a mistake here.

tests/func/test_struct.cpp
```diff
@@ -15,4 +15,28 @@
...
+  EXPECT_EQ(0.5, nd::field_access(nd::field_access(s2, "s1"), "y"));
+
+  nd::array b = nd::array({{1,2,3,4}, {5,6,7,8}, {9,10,11,12}});
+  nd::array s3 = nd::as_struct({{"b", b}, {"s2", s2}});
+  EXPECT_EQ(b, nd::field_access(s3, "b"));
+  //EXPECT_EQ(s2, nd::field_access(s3, "s2"));
+  EXPECT_EQ(7, nd::field_access(s3, "b")(1)(2));
+  EXPECT_EQ(10, nd::field_access(nd::field_access(s3, "s2"), "a")(9));
+}
+
```
izaid: In general, can you use `EXPECT_ARRAY_EQ` instead of `EXPECT_EQ`? The reason being `EXPECT_ARRAY_EQ` will also check the `ndt::type`, whereas `EXPECT_EQ` will only check the values. You shouldn't need to change anything else, just the macro name.

include/dynd/kernels/field_access_kernel.hpp
```diff
@@ -39,7 +71,7 @@ namespace ndt {
 
   template <>
   struct traits<nd::field_access_kernel> {
-    static type equivalent() { return type("({...}) -> void"); }
+    static type equivalent() { return type("({...}, field_name : string) -> Any"); }
   };
```
izaid: A bit of bikesheding here, but what should this callable be called? Do people like `nd::field_access`? Ping @mwiebe @insertinterestingnamehere

tests/func/test_struct.cpp
izaid: Can you also add some tests for arrays of structs? e.g. `10 * {x: float64, y: int32}` and `10 * var * {x: float64, y: int32}` and whatever. This should work just with `elwise`, but we should test it.

skrah: Thanks for all the comments!  -- I still get broadcast errors even with ```EXPECT_ARRAY_EQ(s2, nd::field_access(s3, "s2"));```.  I'll have to look into that.

PR #942, opened 2016-02-15
Test case for broadcast error (struct comparison).

skrah:
I'm not sure, but I think this test -- which gives a broadcast error -- should work.


skrah:
Indeed EXPECT_ARRAY_EQ() works here.

PR #951, opened 2016-02-17
WIP: String split kernel

mdboom:
This just implements splitting on literal strings.  A commonly used feature in Python stdlib is to split on any contiguous whitespace, which this doesn't currently support, and is complicated if you want to support Unicode properly on utf-8 strings (since some whitespace characters are at codepoints > 127).  We should probably clarify the needed feature set in that regard, because it may make the most sense in the long run to just hook into a third-party regex engine to do that kind of thing.  But this PR is still probably "good enough for now" without that and serves as a proof of concept of a string kernel returning var dim arrays.

Some questions:

The general approach here is to allocate the space for the var dim array (using its allocator function) which as far as I can tell gives uninitialized memory, and then I fill it with strings using "placement new".  That's probably the most efficient way (so you don't have to initialize all the resulting strings only to reinitialize them later), but I don't know if there's a better way to spell it -- I find "placement new" isn't always the most well-known feature.

Since this function needs to allocate a var dim array, and getting the allocator from the array metadata is pretty convoluted and tied into kernels etc. I wasn't sure how to also provide the basic function to just do the splitting, as I've done for the other string algorithms.

I couldn't figure out how to write the test to compare the entire var dim array at once, so I'm just testing it element-by-element (this seems to be what the tests in `test_var_dim_type.cpp` do as well, but I don't know if there's a better way now...)

### DIFF
--- a/include/dynd/kernels/string_split_kernel.hpp
+++ b/include/dynd/kernels/string_split_kernel.hpp
@@ -1,0 +1,82 @@
+//
+// Copyright (C) 2011-15 DyND Developers
+// BSD 2-Clause License, see LICENSE.txt
+//
+
+// String split kernel
+
+#pragma once
+
+#include <dynd/string.hpp>
+#include <dynd/string_search.hpp>
+#include <dynd/types/var_dim_type.hpp>
+
+namespace dynd {
+  namespace nd {
+
+    struct string_split_kernel
+      : base_kernel<string_split_kernel, 2> {
+
+      typedef string_split_kernel self_type;
+
+      memory_block_data *m_dst_memblock;
+
+      string_split_kernel(memory_block_data *dst_memblock) :
+        m_dst_memblock(dst_memblock) {
+
+      }
+
+      static void instantiate(char *DYND_UNUSED(static_data), char *DYND_UNUSED(data),
+                              kernel_builder *ckb, const ndt::type &DYND_UNUSED(dst_tp),
+                              const char *dst_arrmeta, intptr_t DYND_UNUSED(nsrc),
+                              const ndt::type *DYND_UNUSED(src_tp),
+                              const char *const *DYND_UNUSED(src_arrmeta),
+                              kernel_request_t kernreq, intptr_t DYND_UNUSED(kwd),
+                              const nd::array *DYND_UNUSED(kwds),
+                              const std::map<std::string, ndt::type> &DYND_UNUSED(tp_vars)) {
+        ckb->emplace_back<self_type>(
+          kernreq, reinterpret_cast<const ndt::var_dim_type::metadata_type *>(dst_arrmeta)->blockref.get());
+      }
+
+      void single(char *dst, char *const *src) {
+        ndt::var_dim_type::data_type *dst_v = reinterpret_cast<ndt::var_dim_type::data_type *>(dst);
+
+        const string *const *s = reinterpret_cast<const string *const *>(src);
+        const string &haystack = *(s[0]);
+        const string &needle = *(s[1]);
+
+        intptr_t count = dynd::string_count(haystack, needle);
+
+        if (count == 0) {
+          dst_v->begin = m_dst_memblock->alloc(1);
+          dst_v->size = 1;
+          string *dst_str = reinterpret_cast<string *>(dst_v->begin);
+          new(&dst_str[0]) string(haystack);
+          return;
+        }
+
+        dst_v->begin = m_dst_memblock->alloc(count + 1);
+        dst_v->size = count + 1;
+        string *dst_str = reinterpret_cast<string *>(dst_v->begin);
+
+        dynd::detail::string_splitter<string> f(dst_str, haystack, needle);
+
+        f(haystack, needle);
+        f.finish();
+      }
+    };
+
+  } // namespace nd
+
+  namespace ndt {
+
+    template<>
+    struct traits<dynd::nd::string_split_kernel> {
+      static type equivalent() {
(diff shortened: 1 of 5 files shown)


include/dynd/kernels/string_split_kernel.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <dynd/types/var_dim_type.hpp>
+
+namespace dynd {
+  namespace nd {
+
+    struct string_split_kernel
+      : base_kernel<string_split_kernel, 2> {
+
+      typedef string_split_kernel self_type;
+
```
izaid: Can we remove the self_type here? That was an old pattern of ours, but is pretty unnecessary now.

include/dynd/kernels/string_split_kernel.hpp
```diff
@@ -0,0 +1,82 @@
...
+      memory_block_data *m_dst_memblock;
+
```
izaid: Rather then a `memory_block_data *`, let's store this as a `intrusive_ptr<memory_block_data>`, which is exactly what it is in the var_dim arrmeta. That gives us reference counting, but the semantics should otherwise be the same.

include/dynd/kernels/string_split_kernel.hpp
```diff
@@ -0,0 +1,82 @@
...
+                              const char *dst_arrmeta, intptr_t DYND_UNUSED(nsrc),
+                              const ndt::type *DYND_UNUSED(src_tp),
+                              const char *const *DYND_UNUSED(src_arrmeta),
+                              kernel_request_t kernreq, intptr_t DYND_UNUSED(kwd),
+                              const nd::array *DYND_UNUSED(kwds),
+                              const std::map<std::string, ndt::type> &DYND_UNUSED(tp_vars)) {
+        ckb->emplace_back<self_type>(
+          kernreq, reinterpret_cast<const ndt::var_dim_type::metadata_type *>(dst_arrmeta)->blockref.get());
+      }
+
```
izaid: You should be able to drop the `.get()` if we change how we store the memblock.

include/dynd/kernels/string_split_kernel.hpp
```diff
@@ -0,0 +1,82 @@
...
+      void single(char *dst, char *const *src) {
+        ndt::var_dim_type::data_type *dst_v = reinterpret_cast<ndt::var_dim_type::data_type *>(dst);
+
+        const string *const *s = reinterpret_cast<const string *const *>(src);
+        const string &haystack = *(s[0]);
+        const string &needle = *(s[1]);
+
+        intptr_t count = dynd::string_count(haystack, needle);
+
```
izaid: Should be able to just do `string_count`, not `dynd::string_count`.

include/dynd/kernels/string_split_kernel.hpp
```diff
@@ -0,0 +1,82 @@
...
+        if (count == 0) {
+          dst_v->begin = m_dst_memblock->alloc(1);
+          dst_v->size = 1;
+          string *dst_str = reinterpret_cast<string *>(dst_v->begin);
+          new(&dst_str[0]) string(haystack);
+          return;
```
izaid: You shouldn't need in-place new here, I believe dst_memblock->alloc should initialize the strings to their default state (a NULL pointer). So you should be able to just do dst_str->assign(...) or whatever.

tests/types/test_string_type.cpp
```diff
@@ -559,6 +559,43 @@ TEST(StringType, Replace) {
...
+  EXPECT_EQ(4, c(2).get_shape()[0]);
+  EXPECT_EQ("c", c(2)(0));
+  EXPECT_EQ("", c(2)(1));
+  EXPECT_EQ("", c(2)(2));
+  EXPECT_EQ("c", c(2)(3));
+
+  EXPECT_EQ(1u, c(3).get_shape().size());
+  EXPECT_EQ(1, c(3).get_shape()[0]);
+  EXPECT_EQ("foobar", c(3)(0));
+}
```
izaid: Okay, so you should be able to use EXPECT_ARRAY_EQ here. It might be a little fiddly with the var dims though, so I'll play with it after we merge this PR.

izaid:
Okay, so hopefully I answered your questions but let me summarize just in case:

1) The memory block allocators should default-construct their elements. Do you shouldn't need placement new, just operate on them as normal.

2) Yeah, we can't really provide this as a basic function as easily. That's fine.

3) It should be possible to simplify the tests, but I need to check some of the details. Don't worry about it.

include/dynd/kernels/string_split_kernel.hpp
mdboom: There is an ambiguity between the regular function in `dynd::string_count` and the elwise kernel in `dynd::nd::string_count`.  In this case I wan't to use the former, but C++ defaults to the latter, since this block is in `dynd::nd`.  I can rename things if that would help...

izaid: Ah, okay, that's cool then.

include/dynd/kernels/string_split_kernel.hpp
mdboom: Ah, indeed.  I wonder why that didn't work for me earlier (I was getting a crash where it was trying to free a garbage pointer when the new string content was assigned).  Anyway, it's working now.

izaid:
@mdboom Sorry, would you mind resolving the merge conflicts? Merging this PR got delayed a bit, but am ready now.

mdboom:
Rebased on master.

izaid:
Thanks, excellent work!

PR #954, opened 2016-02-17
Field Access

izaid:
This implements an `nd::field_access` callable, which is largely @skrah's work.


include/dynd/kernels/field_access_kernel.hpp
```diff
@@ -6,30 +6,62 @@
...
+      static const array error_mode(opt<assign_error_mode>());
+      assign::get()->instantiate(assign::get()->static_data(), data, ckb, dst_tp, dst_arrmeta, 1, field_type,
+                                 field_metadata, kernreq | kernel_request_data_only, 1, &error_mode, tp_vars);
+    };
+
+    static const ndt::type &get_field_type(const ndt::type *src_tp, const array *kwds)
+    {
+      const ndt::struct_type *s = src_tp->extended<ndt::struct_type>();
+      const std::string &name = kwds[0].as<std::string>();
+      uintptr_t index = s->get_field_index(name);
```
skrah: Looking at it again, I think we need something like:
+      intptr_t index = s->get_field_index(name);
+
+      if (index == -1) {
+        throw std::invalid_argument("no field named '" + name + "'");
+      }

Or should we throw the exception in get_field_index()?

izaid: Yeah, you're right, we need an error check like that. I'll add it in after I fix the other issues.

PR #952, opened 2016-02-17
clang-format string-related files

mdboom:
This is a formatting-only change to fixup the string-related stuff I've been working on recently.

(This will need to be done to #951 as well after this is merged).

PR #964, opened 2016-02-19
Add reverse find string kernel

mdboom:
This adds the `rfind` string kernel.

This includes a bit of a refactor of the string search-related algorithms so that the bloom filter search and the operators on that search are combined using function composition and not inheritance.

I probably  could have done "reverse search" with the same code as "forward search" using some iterator tricks, but that got hairier and harder to understand than just reimplementing "reverse search" as a separate function.

This is based on #952, which should be merged first.

### DIFF
--- a/include/dynd/kernels/string_rfind_kernel.hpp
+++ b/include/dynd/kernels/string_rfind_kernel.hpp
@@ -27,7 +27,7 @@ namespace nd {
 namespace ndt {
 
   template <>
-  struct traits<dynd::nd::string_rfind_kernel> {
+  struct traits<nd::string_rfind_kernel> {
     static type equivalent()
     {
       return callable_type::make(ndt::make_type<intptr_t>(), {type(string_id), type(string_id)});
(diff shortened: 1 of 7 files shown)


mdboom:
I've rebased this and fixed things to work with upstream changes.

izaid:
Thank you, will take a look!

include/dynd/kernels/string_rfind_kernel.hpp
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+} // namespace nd
+
+namespace ndt {
+
+  template <>
+  struct traits<dynd::nd::string_rfind_kernel> {
+    static type equivalent()
+    {
```
izaid: Can drop the `dynd::` here.

tests/types/test_string_type.cpp
```diff
@@ -461,8 +461,18 @@ TEST(StringType, Find3)
...
+  nd::array a, b;
+
+  a = {"abc", "ababc", "abcdabc", "abd"};
+  b = "abc";
+  intptr_t c[] = {0, 2, 4, -1};
+
+  EXPECT_ARRAY_EQ(nd::array(c),
+                  nd::string_rfind(a, b));
+}
+
```
izaid: I don't think you need the explicit `nd::array` in EXPECT_ARRAY_EQ.

izaid:
Looks generally very good -- thanks! Happy to merge, more or less as is, but I wanted to leave comments for you for next time.

mdboom:
No worries -- appreciate the comments.  I'll address.

PR #966, opened 2016-02-19
[WIP] Serialization for data described by ndt::type

izaid:


src/dynd/types/bytes_type.cpp
```diff
@@ -50,11 +48,14 @@ void ndt::bytes_type::set_bytes_data(const char *DYND_UNUSED(arrmeta), char *dat
 
 void ndt::bytes_type::print_data(std::ostream &o, const char *DYND_UNUSED(arrmeta), const char *data) const
 {
-  const char *begin = reinterpret_cast<const char *const *>(data)[0];
-  const char *end = reinterpret_cast<const char *const *>(data)[1];
-
-  // Print as hexadecimal
-  hexadecimal_print_summarized(o, begin, end - begin, 80);
+  if (reinterpret_cast<const bytes *>(data)->empty()) {
+    o << "NULL";
```
mwiebe: One thing to ensure is that NULL and a zero-length bytes are distinct values?

izaid: Does that matter to us? How do we want to define `empty`?

mwiebe: I guess I was thinking of NA, NA and zero-length bytes need to be distinct.

Maybe printing with delimiters around the bytes might distinguish the zero-sized bytes? NULL doesn't seem like a zero-length bytes to me. "<>, <03f2>" for example?

PR #970, opened 2016-02-20
Remove array properties from types.

skrah:
This removes all array properties from the types.

### DIFF
--- a/src/dynd/array.cpp
+++ b/src/dynd/array.cpp
@@ -6,10 +6,13 @@
 #include <dynd/array.hpp>
 #include <dynd/array_iter.hpp>
 #include <dynd/callable.hpp>
+#include <dynd/callable_registry.hpp>
+#include <dynd/func/complex.hpp>
 #include <dynd/func/assignment.hpp>
 #include <dynd/func/comparison.hpp>
 #include <dynd/func/elwise.hpp>
 #include <dynd/option.hpp>
+#include <dynd/struct.hpp>
 #include <dynd/types/var_dim_type.hpp>
 #include <dynd/types/fixed_dim_type.hpp>
 #include <dynd/types/tuple_type.hpp>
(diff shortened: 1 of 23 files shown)


src/dynd/array.cpp
```diff
@@ -224,22 +228,77 @@ void nd::array::flag_as_immutable()
...
-    if (!p.is_null()) {
-      return p(eval()); // ToDo: Replace eval() here with an implicit cast
+    const ndt::type &dt = get_dtype();
+
+    switch (dt.get_id()) {
+    case complex_float32_id:
+    case complex_float64_id:
+      return callable_registry[name](eval());
+    case struct_id:
+      return nd::make_field_access_kernel(dt, name)(eval());
```
skrah: This is of course inefficient. We could memoize this later in callable_registry.cpp when we update the kernel.

src/dynd/types/categorical_type.cpp
```diff
@@ -469,19 +469,6 @@ struct get_ints_kernel : nd::base_kernel<get_ints_kernel> {
   }
 };
 
-static const std::map<std::string, nd::callable> &categorical_array_properties()
```
skrah: Apparently there are no tests for this feature. Should this be handled in array.cpp like for structs?

src/dynd/array.cpp
izaid: Yep, noted, but fine for now. :)

src/dynd/types/categorical_type.cpp
izaid: Categorical types are broken, right now. They were too integrated with a variety of things we wanted to remove, so I disabled them. Meaning don't worry for now.

izaid:
For some reason, Travis CI didn't run here. Can you do another push to force it to run?

If the tests pass, I'm happy to merge, but I'll probably do a little follow-up clean-up.

skrah:
Thanks for the review!  I've pushed an unrelated comment change to trigger the build.

izaid:
Merged, thanks!

PR #971, opened 2016-02-22
Remove array functions from types.

skrah:
This removes all array functions from the types. It should be applied on top of #970.


izaid:
Also looks good, merging!

Issue #415, opened 2015-06-27
build error on macosx


jreback:
haven't built in a while, so this could somewhat old

```
(dynd)bash-3.2$ c++ -v
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
```

```
[ 28%] Building CXX object CMakeFiles/libdynd_OBJ.dir/src/dynd/func/arithmetic.cpp.o
/usr/bin/c++    -O3 -std=c++11 -fomit-frame-pointer -fstrict-aliasing -Wall -Wextra -Werror -Wno-missing-field-initializers -fPIC -ferror-limit=20 -O2 -g -DNDEBUG -I/Users/jreback/libdynd/thirdparty/datetime/include -I/Users/jreback/libdynd/include -I/Users/jreback/libdynd/thirdparty/utf8/source    -o CMakeFiles/libdynd_OBJ.dir/src/dynd/func/arithmetic.cpp.o -c /Users/jreback/libdynd/src/dynd/func/arithmetic.cpp
In file included from /Users/jreback/libdynd/src/dynd/func/arithmetic.cpp:6:
In file included from /Users/jreback/libdynd/include/dynd/func/arithmetic.hpp:6:
In file included from /Users/jreback/libdynd/include/dynd/func/arrfunc.hpp:10:
In file included from /Users/jreback/libdynd/include/dynd/config.hpp:142:
/Users/jreback/libdynd/include/dynd/type_sequence.hpp:377:22: error: implicit instantiation of undefined template 'dynd::detail::get_impl<0>'
    typedef typename get_impl<I - 1, A...>::result_type result_type;
                     ^
/Users/jreback/libdynd/include/dynd/type_sequence.hpp:386:18: note: in instantiation of template class 'dynd::detail::get_impl<1, std::__1::pair<const std::__1::pair<dynd::type_id_t, dynd::type_id_t>, dynd::nd::arrfunc> &>'
      requested here
typename detail::get_impl<I, A...>::result_type get(A &&... a)
                 ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/utility:350:86: note: while substituting deduced template arguments into function template 'get' [with I = 1, A =
      <std::__1::pair<const std::__1::pair<dynd::type_id_t, dynd::type_id_t>, dynd::nd::arrfunc> &>]
[...]

PR #973, opened 2016-02-22
Workaround for segfault in Python bindings with clang on Linux

insertinterestingnamehere:
This introduces new declfunc template to demonstrate a possible workaround for the recurring segfault  with clang on Linux that was caused by symbol duplication between libdynd and client libraries under certain circumstances.

My current best theory is that the static data at https://github.com/libdynd/libdynd/blob/d14ee985052c14c08ffa156c7e70dff0cc3dcfc4/include/dynd/callable.hpp#L340 is getting inadvertently duplicated. Looking at the symbol tables it's already clear that the get method is getting instantiated wherever it's used, so it's not overly surprising that the static data gets dragged along with it. I haven't yet been able to pin down what circumstances make the duplication start happening. It'd be nice to pin that down better so we could submit a bug report upstream.

This PR, at least for now, shows a new idiom that we can use as a replacement for the current declfunc template that avoids these static initialization issues by forcing the get method of the structs created that way to be linked from libdynd. I've only applied it in the relevant place for assignment thus far. Feedback would be appreciated before I apply the replacement everywhere.

### DIFF
--- a/include/dynd/callable.hpp
+++ b/include/dynd/callable.hpp
@@ -7,15 +7,15 @@
 
 #include <memory>
 
+#include <dynd/callables/static_data_callable.hpp>
 #include <dynd/config.hpp>
-#include <dynd/kernels/apply_function_kernel.hpp>
 #include <dynd/kernels/apply_callable_kernel.hpp>
+#include <dynd/kernels/apply_function_kernel.hpp>
 #include <dynd/kernels/apply_member_function_kernel.hpp>
 #include <dynd/kernels/construct_then_apply_callable_kernel.hpp>
 #include <dynd/types/callable_type.hpp>
-#include <dynd/types/substitute_typevars.hpp>
 #include <dynd/types/option_type.hpp>
-#include <dynd/callables/static_data_callable.hpp>
+#include <dynd/types/substitute_typevars.hpp>
 
 namespace dynd {
 namespace nd {
@@ -343,6 +343,27 @@ namespace nd {
   };
 
   template <typename FuncType>
+  struct declfunc2 {
+    template <typename DstType, typename... ArgTypes>
+    array operator()(ArgTypes &&... args)
+    {
+      return FuncType::get().template operator()<DstType>(std::forward<ArgTypes>(args)...);
+    }
+
+    template <typename... ArgTypes>
+    array operator()(ArgTypes &&... args)
+    {
+      return FuncType::get()(std::forward<ArgTypes>(args)...);
+    }
+
+    array operator()(const std::initializer_list<array> &args,
+                     const std::initializer_list<std::pair<const char *, array>> &kwds)
+    {
+      return FuncType::get()(args, kwds);
+    }
+  };
+
+  template <typename FuncType>
   std::ostream &operator<<(std::ostream &o, const declfunc<FuncType> &DYND_UNUSED(rhs))
   {
     return o << FuncType::get();
@@ -466,8 +487,7 @@ namespace nd {
   } // namespace dynd::nd::functional
 
   template <typename CallableType, typename... T, typename>
-  callable::callable(CallableType f, T &&... names)
-      : callable(nd::functional::apply(f, std::forward<T>(names)...))
+  callable::callable(CallableType f, T &&... names) : callable(nd::functional::apply(f, std::forward<T>(names)...))
   {
   }
 
--- a/include/dynd/func/assignment.hpp
+++ b/include/dynd/func/assignment.hpp
@@ -11,8 +11,9 @@
 namespace dynd {
 namespace nd {
 
-  extern DYND_API struct DYND_API assign : declfunc<assign> {
+  extern DYND_API struct DYND_API assign : declfunc2<assign> {
     static callable make();
+    static callable &get();
   } assign;
 
 } // namespace dynd::nd
--- a/src/dynd/func/assignment.cpp
+++ b/src/dynd/func/assignment.cpp
@@ -24,8 +24,8 @@ struct DYND_API _bind {
 DYND_API nd::callable nd::assign::make()
 {
   typedef type_id_sequence<bool_id, int8_id, int16_id, int32_id, int64_id, int128_id, uint8_id, uint16_id, uint32_id,
-                           uint64_id, uint128_id, float32_id, float64_id, complex_float32_id,
-                           complex_float64_id> numeric_ids;
+                           uint64_id, uint128_id, float32_id, float64_id, complex_float32_id, complex_float64_id>
+      numeric_ids;
 
   ndt::type self_tp = ndt::callable_type::make(ndt::any_kind_type::make(), {ndt::any_kind_type::make()}, {"error_mode"},
                                                {ndt::make_type<ndt::option_type>(ndt::make_type<assign_error_mode>())});
@@ -119,6 +119,12 @@ DYND_API nd::callable nd::assign::make()
     }
     return child;
   });
+}
+
+DYND_API nd::callable &nd::assign::get()
+{
+  static nd::callable self = nd::assign::make();
+  return self;
 }
 
 DYND_API struct nd::assign nd::assign;


insertinterestingnamehere:
In the long term I'd like to get rid of the declfunc template entirely. It exists primarily for lazy initialization and we don't currently have much structure that actually requires that. I don't know if pulling it out right now is the best thing just yet though.

izaid:
Okay, so let me split my comments into two parts.

First, if this fixes the clang bug, then we should merge this and accept we need some special `declfunc2` to keep `nd::assign` working for now.

Second, in principle, I'd like to do away with `declfunc` as well. But I'm not sure this is possible. In C++, we have no control over the order of static initialization. So if someone tries to make a callable that uses a statically-declared `ndt::type` or another statically-declared `nd::callable`, the only way to make that work is with explicit initialization or lazy initialization. Even if we sorted out the `ndt::type` parts, I can imagine someone wanting to make a `nd::callable` that uses a pre-existing DyND callable in some way (`nd::assign` or `nd::add` or whatever). Without lazy initialization, there's no guarantee that will work (and it probably won't).

So I think we need to accept we need lazy initialization. That doesn't preclude a massive refactoring of declfunc. @insertinterestingnamehere, what do you want to see happen?

insertinterestingnamehere:
In the long run, I'd like it if we could make things like `nd::assign`, `nd::add`, and whatever else plain and simple callables that are build from their corresponding make functions at load time and, rather than providing an idiom for lazy initialization, we provided an interface for guaranteeing that a given callable has already been initialized when constructing something else. The initialization function would have essentially the same purpose as our current `declfunc::get` method does, except that things would be initialized entirely at load time rather than on first call and each callable's make function would have to explicitly call the make functions for other callables it's borrowing kernels from. In the vast majority of cases that I'm aware of we wouldn't need to do anything other than write the needed make function. and let the compiler initialize things for us. In the few cases where there were interdependencies, we could use a different idiom to control initialization order at load time.

That said, for now, since the symbol duplication isn't unique to `nd::assign` I'd vote for replacing declfunc entirely with the newer version in this PR. That at least makes our dynamic overloading work right for all our current callables.

insertinterestingnamehere:
An explicit example would probably be more clear. Here's roughly what I'd like to see us doing, though there are probably still a few things to sort out.

    nd::callable make_callable_1() {
      // do initialization stuff for the first callable
    }
    nd::callable &get_callable_1() {
      static nd::callable self = make_callable_1();
      return self;
    }
    nd::callable &callable_1 = get_callable_1();
    nd::callable make_callable_2() {
      local_callable_1 = get_callable_1();
      // More initialization stuff
    }
    nd::callable &get_callable_2() {
      static nd::callable self = make_callable_2();
      return self;
    }
    nd::callable &callable_2 = get_callable_2();

I'm sure there's some way to do a similar initialization pattern without having to export an lvalue reference, but this is good enough to at least demonstrate the idea.

insertinterestingnamehere:
Test failure is some sort of timeout issue. I'll restart the relevant build.

skrah:
Thanks for fixing the segfault!  -- I'm still getting a broadcast error here (Ubuntu Linux, clang++-3.6, stdlib=libc++):

$ ./dbg/bin/python3  -c "import dynd; dynd.test(verbosity=2, exit=True)"
Running unit tests for the DyND Python bindings
Python version: 3.6.0a0 (default:cbf83f55d5f3, Feb 23 2016, 12:04:43) 
[GCC 4.2.1 Compatible Ubuntu Clang 3.6.0 (tags/RELEASE_360/final)]
Python prefix: /home/stefan/dbg
DyND-Python module: /home/stefan/dbg/lib/python3.6/site-packages/dynd-0.6.7.dev1267-py3.6-linux-x86_64.egg/dynd
DyND-Python version: b'v0.6.6-1267-g5502865-dirty'
DyND-Python git sha1: b'55028657e27ecccf48c108d3451ef51ca04d7dca'
LibDyND version: b'v0.6.6-3073-g6446fe1-dirty'
LibDyND git sha1: b'6446fe132b062f0b973b89a15e07537fddd75b8f'
NumPy version: 1.11.0.dev0+3af5f05
test_pyconvert (dynd.tests.test_int128.TestInt128) ... ok
test_pyconvert_overflow (dynd.tests.test_int128.TestInt128) ... ok
test_pyconvert (dynd.tests.test_int128.TestUInt128) ... ok
test_pyconvert_overflow (dynd.tests.test_int128.TestUInt128) ... ok
test_numba (dynd.tests.test_functional.TestApply) ... skipped "No module named 'numba'"
test_object (dynd.tests.test_functional.TestApply) ... ok
test_unary (dynd.tests.test_functional.TestElwise) ... ok
test_unary (dynd.tests.test_functional.TestReduction) ... ok
test_binary (dynd.tests.test_annotate.TestAnnotate) ... ok
test_binary_with_return (dynd.tests.test_annotate.TestAnnotate) ... ok
test_nullary (dynd.tests.test_annotate.TestAnnotate) ... ok
test_nullary_with_return (dynd.tests.test_annotate.TestAnnotate) ... ok
test_unary (dynd.tests.test_annotate.TestAnnotate) ... ok
test_unary_with_return (dynd.tests.test_annotate.TestAnnotate) ... ok
test_struct_or_tuple (dynd.tests.test_array_as_py.TestArrayAsPy) ... ok
test_pickle_types (dynd.tests.test_pickle.TestPickle) ... ok
test_object_arr (dynd.tests.test_array_copy.TestCopyFromNumPy) ... ok
test_object_in_struct_arr (dynd.tests.test_array_copy.TestCopyFromNumPy) ... ok
test_simple_strided (dynd.tests.test_array_copy.TestCopyFromNumPy) ... ok
test_simple_var (dynd.tests.test_array_copy.TestCopyFromNumPy) ... ok
test_bool (dynd.tests.test_array_copy.TestCopyFromPy) ... ok
test_simple_fixed_dim (dynd.tests.test_array_copy.TestIteratorAssign) ... terminating with uncaught exception of type dynd::broadcast_error
Aborted (core dumped)

insertinterestingnamehere:
@skrah I'm intrigued libdynd is building with that configuration at all. Where did you get your LLVM toolchain? Only recent versions of libc++ have avoided https://github.com/libdynd/libdynd/issues/415 on Linux. I could see the LLVM apt packages or the packages for ubuntu 15.10 having a recent enough version.

skrah:
Let me see. I /think/ that all packages that are used are from Ubuntu:

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.4 LTS
Release:        14.04
Codename:       trusty

$ dpkg --get-selections | grep llvm    
libllvm3.2:amd64                                deinstall
libllvm3.4:amd64                                install
libllvm3.6:amd64                                install
llvm-3.4-runtime                                install
llvm-3.6                                        install
llvm-3.6-dev                                    install
llvm-3.6-doc                                    install
llvm-3.6-examples                               install
llvm-3.6-runtime                                install
llvm-3.6-tools                                  install
llvm-runtime                                    install

$ dpkg --get-selections | grep clang   
clang-3.6                                       install
clang-3.6-doc                                   install
clang-3.6-examples                              install
clang-format-3.6                                install
libclang-common-3.6-dev                         install
libclang1-3.6:amd64                             install
libclang1-3.6-dbg                               install

I'm building with:
export CC=clang &&
export CXX=clang++ &&
export CXXFLAGS="-stdlib=libc++"

mkdir build &&
cd build &&
cmake -DCMAKE_BUILD_TYPE=debug ..


Now, I *might* have compiled llvm-3.6 and clobbered the distro packages at some point, but I don't think so. It's a mystery. :)

skrah:
In #415 -std=c++11 is used.  That did not work for me either -- clang seems to need  -std=c++14.

insertinterestingnamehere:
That's awesome. Thanks for finding that. I guess we hadn't checked that bug since upgrading to C++14. I'll take a look and see how everything's doing there. I'm working on other projects most of the day today, but I'll be able to do that tomorrow.

izaid:
+1 to both of you being awesome and sorting this out.

PR #974, opened 2016-02-23
Remove type functions.

skrah:
I think the type functions were entirely unused. This just removes them all.


izaid:
Looks good, merged!

PR #979, opened 2016-02-24
WIP: Remove callables from type properties.

skrah:
This is a union-based approach to type properties. I thought it might be clearer, since the consumer of the properties needs to switch on some kind of id/tag anyway.


include/dynd/types/type_id.hpp
```diff
@@ -175,6 +175,35 @@ enum type_id_t {
...
+  Uintptr_kind,
+  String_kind,
+  Type_kind,
+  UintptrVector_kind,
+  StringVector_kind,
+  TypeVector_kind
+};
+
+typedef struct {
+  type_property_kind_t kind;
```
skrah: Here we could in fact use type_id_t, but: Some types are missing, and we would need a reserved flag (something like the high bit of a size_t, if the type_id_t enum is changed to size_t) to distinguish between vectors and plain types.

skrah:
Clang accepts this ```properties["return_type"] = {.kind = Type_kind, {.type = &m_return_type}};``` here, but the other compilers fail (apparently this is C99 only).

izaid:
Okay, so my overall thoughts are that this approach looks good! In general, I like the idea of a simple `type_property_t` tagged union more than the `char *` approach. A few thoughts:

1) We can't have `std::vector<T>` as a possibility here, because that's not compatible with a C ABI. Anything that uses a `std::vector<T>` needs to have it's type property be a `T *` and a `size_t` for the size. I'm not sure what the right approach is for that, but if we can include it in the union that's great.

2) Similar thoughts apply for `std::string`, we need to replace that with a `char *`.

3) I'm very against introducing another enum (namely `type_property_kind_t`) into DyND. If we rethink how we are doing the vectors, everything should be covered by `type_id_t`. For something like `intptr_t` or `size_t`, we can add aliases like `intptr_id` or `size_id` that map to the corresponding existing type IDs. Actually, something like `type_id_of<intptr_t>::value` should work right now.

izaid:
Technically, `properties["return_type"] = {.kind = Type_kind, {.type = &m_return_type}};` should work, as I thought C99 was supposed to be totally valid in C++11.

izaid:
Pinging @insertinterestingnamehere and @mwiebe here as well, so they can voice whether or not they like the tagged union approach.

insertinterestingnamehere:
I still favor the approach using types that we discussed yesterday. We already have a type system to represent stuff like this. It doesn't make any sense in my mind to replace it with something more restrictive. That said, this is stretching our interface between DyND types and C++ types quite a bit, so any solution would be good for now.

izaid:
@insertinterestingnamehere But, if we use `type_id_t` from DyND as the union ID then we are just using a subset of our existing type system. Do you think we need access to anything that can `ndt::type` can represent? That makes copying to/from the `char *` quite tricky.

insertinterestingnamehere:
I'm not aware of any reasons right now to require types for this. That said, in the long run, though, for user defined types and in terms of a user interface, it'd be nice if we re-used our type system for this. The fact that we haven't settled the `char*` casting issue means we can't really do that in the sort term, but I'd still really like to do that in the long run.

skrah:
In theory we could put a ```* void``` in the union that is used if a custom type sets its own dynamic type id as the tag.  But how to interpret that unless the custom type is known in advance?  Actually I have the same question about  the ```* char``` proposal we discussed yesterday.

insertinterestingnamehere:
The only way to allow mapping from dynamic types back to static types (that I'm aware of) is to have users specify the type they expect, map the provided static type to its corresponding dynamic type and then do a runtime check that it matches the dynamic type they are given. Matching can be taken to mean a variety of things. Exact matching, allowing subclasses, and allowing casts are all viable options. It's not optimal from a performance perspective, but it's the best we can do with the dynamic type system we have.

mwiebe:
Would it be reasonable to return an ndt::type object as part of the property access, instead of adding a new enum? Where the nd::array is available, this can very conveniently be wrapped up into one of those, and in other contexts the subset of supported types can be done with a switch statement covering the cases of the enum in this patch.

mwiebe:
To try and phrase this a bit differently: It feels nice to me to use ndt::type to describe the data for ndt::type properties. I think to do it differently needs a strong argument why the other way of describing them is better, and begs the question why the distinguishing aspects don't go into ndt::type itself.

skrah:
I think the main problem is that several properties are values rather than types, e.g. https://github.com/libdynd/libdynd/pull/979/files#diff-6f52797385ab1f82e2f4072c58ebed3cR620 . Would you restrict the properties to returning types only?

mwiebe:
The properties would be typed values, using ndt::type instances to describe what types they are, or equivalently how to interpret the raw pointer returned. If the property itself is a type, its type is ndt::type("type").

insertinterestingnamehere:
Yep, my proposal is to have an api that can be used like this: `const expected_type &value = type.p<expected_type>("property_name");`. I'm still not clear on why we'd need a switch statement somewhere to allow that sort of thing to work though. Why is that?

insertinterestingnamehere:
At a low level that would return a `void*` and an `ndt::type`, do a dynamic check to verify that the type returned from the map matches the type, and then cast to a reference of the desired type.

izaid:
@insertinterestingnamehere Yes, that should work.

skrah:
@insertinterestingnamehere I thought one of the reasons to split off the types was to make life easier for C users. If a C function gets an ```{ ndt::type, void * }``` struct and wants to create appropriate Python objects from that, I don't see how to avoid the switch. -- This is just an explanation why I repeatedly mention switch statements (which would of course also be needed for extracting a tagged union member if the type is dynamic).

But perhaps I misunderstand how the dynamic type properties are used inside dynd-python. For me, a fully dynamic property access function would have a signature like ```(ndt::type, string) -> PyObject```.

skrah:
With the latest proposal: How to represent ```uintptr_t``` and ```size_t```? I think there are IBM systems where ```void *``` is 128 bit and ```size_t``` 64 bit.

insertinterestingnamehere:
Hmmm. Good questions. So, on the Python side we can set up dynamic registries of function pointers that essentially take a value of a given dynd type and convert it to a Python object. The main downside of doing this here is that building a dispatching system and a registry like that is essentially what we're doing with callables already. The only difference here would be that this version would need to operate on `char*` values rather than arrays. Re-creating a minimal callable is quite a tall order when we're trying to just separate library dependencies. I'd really appreciate it if @mwiebe and @izaid could weigh in on this one. The best thing to do short term is probably just to have a switch statement in Python for now until we can figure out how to have something more dynamic there.

insertinterestingnamehere:
And thanks for pointing out that `uintptr_t` and `size_t` are different. I wasn't aware of that. The most natural thing seems to be to just add types for `size_t`, `ssize_t`, `uintptr_t`, and `intptr_t`. I'll let Mark and Irwin weigh in there as well though. It's intriguing that we don't have those types already.

mwiebe:
@skrah The `{ndt::type, void *}` has the following properties which favour it in my mind:

* Supports general DyND types for properties, future extension requires no additional work in this part of DyND.
* Choosing a subset of supported types for a Python binding requires only slightly different switch statement structure than the flat enum approach. It's a switch(type_id), and for fixed_dim_type, a static_cast followed by a second switch(type_id) on the element type.

I don't understand where `uintptr_t` and `size_t` come into play here? They're just platform-specific aliases to particular `uint##` types, aren't they?

mwiebe:
See http://datashape.pydata.org/grammar.html#the-datashape-grammar which includes `uintptr` and `intptr`. Should `size` and `ssize` be added as additional aliases?

skrah:
Superseded now by  #1009.

 @mwiebe: Sorry for answering your questions so late.  In #1009 ```uintptr_t``` etc. should be fine, since there is no need to use something like ```uintptr_id``` or ```size_id``` in a switch statement.  Previously I thought that these ids would be needed.  In #1009 I have added ```size``` as an alias for creating ndt::type("size"),

PR #989, opened 2016-02-26
Removed some duplicate entries in the callable_registry, untangled some of the include files around math.hpp

izaid:


izaid:
Dealt with this in another PR.

PR #1009, opened 2016-03-02
WIP: Remove callables from type properties (alternative version).

skrah:
@insertinterestingnamehere @izaid @mwiebe  

This is the alternative (union-free) version for removing callables from the type properties. I think this is what Ian suggested. It should support all scalar types, strings, ndt::type as well as vectors of these types.

For C++ it is probably reasonable, I'm not sure yet if it is immediately usable in dynd-python.

The union approach also had some problems (poor initialization support in gcc) and would have needed pretty much the same switch statements.

I wasn't sure where to put helper classes like is_vector. They probably need to go somewhere else.


skrah:
I'm not sure what happens on OS X here. Apparently ```std::is_same<size_t, type_of<uint64_id>::type>::value``` fails but works on the other platforms.

include/dynd/type.hpp
```diff
@@ -281,13 +281,99 @@ namespace ndt {
...
+      }
+    }
+
+    template<typename T>
+    bool has_admissible_property_type() const
+    {
+       return has_scalar_type<T>() || has_admissible_value_type<T>();
+    }
+
     /**
```
insertinterestingnamehere: This comment should probably be moved down to where the property access method is.

insertinterestingnamehere:
This is awesome. Thank you! Modulo the comment placement and the issues we've already discussed offline, this looks great to me.

include/dynd/types/type_id.hpp
```diff
@@ -355,6 +355,11 @@ struct type_id_of<void> {
   static const type_id_t value = void_id;
 };
 
+template <>
+struct type_id_of<ndt::type> {
+  static const type_id_t value = type_id;
+};
+
```
skrah: This one could also go into the ```property_type_id_of``` set.  I wasn't sure if it was deliberately missing here or not.

izaid: That was accidentally missing, good fix!

PR #1011, opened 2016-03-06
Clang/Linux: Fix vtable duplication between libdynd and dynd-python DSOs.

skrah:
This fixes the long standing clang/Linux exception handling issue. @insertinterestingnamehere was on the right track (thanks!).

The patch contains many clang-format artifacts, I hope it's still readable.


izaid:
@insertinterestingnamehere Can you take a look here? I'm not familiar enough with this issue.

skrah:
It's pretty much the same as https://codereview.qt-project.org/#/c/144299/ . It looks like the Qt project had several of these issues.

izaid:
From my reading of that, it seems like we're *not* supposed to have virtual destructors defined in *.cpp files. Is that correct?

Ping @mwiebe

skrah:
Yes (I think you meant *.hpp files): Basically no inline virtual destructors in header files.

The patch goes a bit further and moves other definitions into the .cpp file, too.  I did not test the smallest set of changes that would appease clang.

izaid:
Yes, sorry, I meant *.hpp files. That's very interesting.

skrah:
Yes, Graham mentioned in Flowdock that it's debatable whether it's a compiler bug or actually required by the C++ standard.  The problem is that DSO behavior is probably ill-specified (if at all).

insertinterestingnamehere:
I can confirm that this fixes the issue with clang 3.8-rc2 and clang 3.7. Merging.

insertinterestingnamehere:
Thanks for tracking this down. I had tried a few different combinations of non-inlined methods, but hadn't gotten it pinned down this far yet. It's really good to know about potential hangups like this. Thanks!

PR #1013, opened 2016-03-07
make_field_access_kernel(): convert assert() into an exception

skrah:
I just noticed that make_field_access_kernel() needs to check for struct_id if the caller doesn't.


skrah:
Branching went wrong: The actual change is trivial and in ```src/dynd/struct.cpp```.

PR #1022, opened 2016-03-09
Fix invalid access.

skrah:
This fixes a valgrind error in the dynd tests.


include/dynd/parse_util.hpp
```diff
@@ -1056,8 +1056,9 @@ std::enable_if_t<is_floating_point<T>::value, T> parse(const char *begin, const
 
   // TODO: use http://www.netlib.org/fp/dtoa.c
   char *end_ptr;
-  T value = strto<T>(begin, &end_ptr);
-  if (end_ptr - begin != end - begin) {
+  const char *cp = std::string(begin, end).c_str();
```
mwiebe: This will make a full copy of the string, there's got to be a way to fix this which doesn't trigger that.

skrah: We could use a buffer on the stack for regularly sized inputs and fall back to std::string or similar for huge inputs.  I think the memcpy() into a static buffer is probably negligible compared to strtod() and for sure compared to dtoa.c (mentioned in the TODO comment).

mwiebe:
Fixing up dynd::string to always include the nul terminator should fix this for that case, but I guess this issue still exists for fixed_string because we're using NumPy semantics there which doesn't guarantee the nul terminator

mwiebe:
It does really suck making that copy, because this might be the start of an arbitrarily large string buffer.

skrah:
The copy is bad, but there are already several similar copies in that file. Perhaps we can fix this and revisit the efficiency issue later.

PR #1027, opened 2016-03-11
Fix cmake "liblib" problem on Linux.

skrah:
libdyndt.so is currently installed as liblibdyndt.so on Linux.


insertinterestingnamehere:
Good catch. Thanks. I'll merge when the tests pass.

insertinterestingnamehere:
Trying https://github.com/libdynd/libdynd/pull/1028 to see if that fixes the Windows-specific issue here.

PR #1090, opened 2016-03-29
Add deprecated versions of the old elwise and reduction headers.

insertinterestingnamehere:
Given that this is one of the better known parts of our API, I figured it would be good to do an actual deprecation cycle for the old `<dynd/func/elwise.hpp>` and `<dynd/func/reduction.hpp>` headers before removing them in favor of `<dynd/functional.hpp>`. This uses the C++14 deprecated syntax on the most commonly used versions of `dynd::nd::functional::elwise` and `dynd::nd::functional::reduction`. The deprecations are only picked up if the additional headers containing these forward declarations are included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,6 +334,8 @@ set(libdynd_SRC
     include/dynd/diagnostics.hpp
     include/dynd/ensure_immutable_contig.hpp
     include/dynd/fft.hpp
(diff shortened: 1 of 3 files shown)


insertinterestingnamehere:
Also, the new syntax for deprecations in C++14 is fantastic. It makes stuff like this really easy.

izaid:
Looks good to me!

PR #1091, opened 2016-03-29
Braces

insertinterestingnamehere:
This implements my old suggestion of changing the clang-format configuration to attach braces for functions, catch statements, and else statements. The relevant change is one line and is the only thing in the first commit. The second commit applies clang-format to our whole source tree. The third commit  adds various missing includes that clang-format's reordering of include directives exposed.

This is really just a suggestion, take it or leave it. Input is welcome.


insertinterestingnamehere:
There's a net change of -3414 lines here. github's line counter is off on this one.

PR #1092, opened 2016-03-29
Check for assumptions about float sizes in config.hpp.

skrah:
config.hpp assumes sizeof(float)==4 and sizeof(double)==8.  The patch checks this in the ```cmake``` stage.

I think the assumption holds for most systems, so it should be fine.  If this goes in, the ```real``` type is effectively locked down to ```float64``` and ```complex``` to ```complex128```.

Perhaps we should also make that explicit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@
 
 cmake_minimum_required(VERSION 2.8.11)
 project(libdynd)
+include(CheckTypeSize)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/")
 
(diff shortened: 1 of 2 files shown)


izaid:
This is a good point. I've been wondering for a long time whether we should be more careful about float32, float64, and so on. In practice, this won't be a problem on any systems we're currently on, but it's worth thinking about it.

insertinterestingnamehere:
We generally assume IEEE floating point arithmetic with standard floating point sizes. That said, checking for this sort of thing up-front seems like a good idea. This looks good to me.

mwiebe:
Is it reasonable to require that float/double are IEEE 754? I.e. that http://en.cppreference.com/w/cpp/types/numeric_limits/is_iec559 returns true?

insertinterestingnamehere:
I like the idea of checking that. Other floating point types can be supported as other types in our type system. If someone asks for a float32 or a float64, they should get something that is appropriately sized and has the expected IEEE semantics.

skrah:
@mwiebe Some background on this patch:  I'm in the process of documenting datashape *as used in libdynd*.   It would be good like to give some hard guarantees in the docs that e.g. ```complex128``` always consists of two doubles of size 8.

So, for the documentation the ```__STDC_IEC_559__``` issue is not so important right now. The sizes are, since we're also trying to define a way to make callables work on PEP-3118 buffers.

skrah:
Of course, +1 for additionally requiring IEEE semantics.

mwiebe:
I wrote https://github.com/blaze/datashape/blob/master/docs/source/types.rst#binary-floating-point and intended for this to apply. I see there I defined `char` the way we discussed in the other issue, too.

skrah:
I like both of your original specifications. They'll probably work well in practice (and they'd make PEP-3118 interoperability a lot easier).

PR #1094, opened 2016-03-30
Change encoding properties to return strings. Remove latin1 encoding.

skrah:
This changes the encoding properties from integers to strings, so that the dynd-python representation becomes human-readable.

Additionally, this removes the latin1 encoding since it was unsupported by the parser and is not present in blaze datashape.  The alternative would be to make the parser accept latin1.

### DIFF
--- a/include/dynd/types/fixed_string_type.hpp
+++ b/include/dynd/types/fixed_string_type.hpp
@@ -17,7 +17,8 @@ namespace ndt {
 
   class DYNDT_API fixed_string_type : public base_string_type {
     intptr_t m_stringsize;
-    string_encoding_t m_encoding;
+    const string_encoding_t m_encoding;
+    const std::string m_encoding_repr;
 
   public:
     fixed_string_type(intptr_t stringsize, string_encoding_t encoding);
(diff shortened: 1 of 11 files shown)


include/dynd/types/fixed_string_type.hpp
```diff
@@ -17,7 +17,8 @@ namespace ndt {
 
   class DYNDT_API fixed_string_type : public base_string_type {
     intptr_t m_stringsize;
-    string_encoding_t m_encoding;
+    const string_encoding_t m_encoding;
+    const std::string m_encoding_repr;
```
mwiebe: Is there a reason to store a copy of this in here? I think it's probably nicer to just have the enum to keep things lighter weight.

src/dynd/types/base_string_type.cpp
```diff
@@ -23,7 +23,7 @@ size_t ndt::base_string_type::get_iterdata_size(intptr_t DYND_UNUSED(ndim)) cons
 std::map<std::string, std::pair<ndt::type, const char *>> ndt::base_string_type::get_dynamic_type_properties() const
 {
   std::map<std::string, std::pair<ndt::type, const char *>> properties;
-  properties["encoding"] = {ndt::type("uint32"), reinterpret_cast<const char *>(&m_encoding)};
+  properties["encoding"] = {ndt::type("string"), reinterpret_cast<const char *>(&m_encoding_repr)};
```
mwiebe: I see this is why the copy of encoding repr is there.

mwiebe:
I think a general thing we might want to change within dyndt is all `std::string` to `nd::string`, so that string data can be viewed without a copy as DyND data.

PR #1097, opened 2016-03-30
[WIP] Moving to a consistent callable model

izaid:
DyND's callable model evolved organically over time, as @mwiebe and I added things like keywords and per call data. While it is general enough to do pretty much anything, there is misuse of it throughout DyND. This is an attempt to reign in misbehaving callables, notably the functionals, by giving DyND a very consistent call graph of 1 resolution, followed by 1 instantiate.

The basic idea is as follows. Each callable will have a pair of methods called "resolve" and "instantiate". 
The "resolve" method gets access to the types, the keywords, any pattern-matched type variables, and the callable that called it (NULL initially). It is able to deduce a return type from these and initialize some per call data. It also must add any of it's children to a "call graph" structure ONLY IF they are instantiable callables -- i.e., they actually correspond to a kernel, not something like multidispatch. The "instantiate" method gets access to the metadata, the kernel request, and the per call data from resolve. In "instantiate", the call graph already exists, so instantiate must simply walk the metadata tree, creating it's own kernel and the kernels of its children in the call graph.

By structuring DyND callables like this, we have exactly the same abilities as before, but there is a clear separation between the type dispatch and the metadata dispatch. This will naturally allow for two independent specializations, 1 on types and 1 on metadata. It also fixes the long-standing reference cycle problem of DyND's callables.


izaid:
This, I believe, will make things in general more efficient (although that is not the primary concern). What is worth considering is that callables will no longer be able to directly access the types or keywords from instantiate. Most of the time, this doesn't matter. A simple callable doesn't need those things anyway, and the very complex callables (like elwise) are already unpacking them into temporary storage.

But, there are intermediate callables (like those involving strings) that pipe type properties and / or keywords directly into the kernels. This PR will ultimately force them to put them into temporary (binary) storage. So we'll have some extra copies of primitive data types (think a few extra if statements and a couple of integer copies), at the price of consistency.

If this is a problem, there are potential solutions. For the keywords, we can allow instantiate to have access to them still. For the types, we might want to redundantly store certain information in the metadata. We do this for the of a fixed_dim, but perhaps we'd also want to do it for things like string encodings.

The intermediate case I describe above does not really concern me, but it is something that I crossed my mind. For everything else, I think this is an improvement.

izaid:
I should also mention that this is pretty much necessary to fix long-standing bugs in reductions and elwise, among others. Basically, the functionals need to conform to a model to like this to make them maintainable. Elwise currently assumes that it's children never have per call data, whereas reduction is relying on a hard-coded table of 8 function pointers.

izaid:
This PR got a little out of hand, so I'm probably going to try with a different PR in more bite-size pieces. Not totally sure *how* to do bite-sized pieces with something this big, but we'll see...

mwiebe:
I think limiting exposure of the keyword parameters to `resolve` is a good thing. The keyword arguments are dynamic, and it's the job of `resolve` to convert that into something it knows about statically, which it forwards on to the kernel's `instantiate`. If `instantiate` gets the keyword arguments as well, it again has to validate the dynamic type again instead of just grabbing statically typed data.

izaid:
Closing, because we achieved this now.

PR #1106, opened 2016-04-01
Increase MSVC version where workaround to fix dynd::outer applies.

insertinterestingnamehere:
The bug in handling the template patterns used in `dynd::outer` appears to also be present in MSVC 2015 update 2, so I've updated the ifdef guards to use the workaround there as well.

### DIFF
--- a/include/dynd/type_sequence.hpp
+++ b/include/dynd/type_sequence.hpp
@@ -68,8 +68,7 @@ namespace detail {
 } // namespace dynd::detail
 
 template <typename SequenceType>
-const std::array<typename SequenceType::value_type, SequenceType::size> &i2a()
-{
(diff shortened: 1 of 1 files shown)


include/dynd/type_sequence.hpp
```diff
@@ -321,7 +318,7 @@ struct outer;
 
 // Work around bug in MSVC 2015 update 1 and earlier.
```
mwiebe: The comment should say "update 2" now as well

insertinterestingnamehere: Thanks for pointing that out. I'll fix that.

insertinterestingnamehere: Fixed in https://github.com/libdynd/libdynd/commit/2c7fae5b5621dc0227326c6c31b8476fb03d0914.

PR #1110, opened 2016-04-01
Updates to README and other related files.

insertinterestingnamehere:


izaid:
Yeah, this is definitely a solid step in the direction. Thanks!

PR #1125, opened 2016-04-06
Check for existence of arithmetic operators when expanding kernels for different types.

insertinterestingnamehere:


izaid:
@insertinterestingnamehere So, is the basic idea here that we check for the existence of arithmetic operators for any types when we make kernel? If yes, that's awesome and I'm very happy with it.

insertinterestingnamehere:
Yep. That's exactly what this does. There may still be some details to work through with ambiguous operators, but this puts the machinery in place so that we can start letting the compiler automatically include arithmetic operators for our built in types based on whether or not they are defined.

izaid:
Perfect, let me know when you merge and I'll proceed with my clean-up of kernels/arithmetic.hpp

PR #1130, opened 2016-04-07
Workaround compiler bug in MSVC 2015 involving initializer lists

insertinterestingnamehere:
For some reason it's not always letting initializer lists be used with range based for loops. Still no minimal reproducible example. This workaround works, but the problem is pretty clearly a compiler bug.

### DIFF
--- a/include/dynd/callables/forward_na_callable.hpp
+++ b/include/dynd/callables/forward_na_callable.hpp
@@ -30,7 +30,7 @@ namespace nd {
 
         kb(kernel_request_single, dst_arrmeta, nsrc, src_arrmeta);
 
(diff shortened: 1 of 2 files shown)


izaid:
Thanks for sorting that out! I think we need to just come up with a good idiom for that pattern that all compilers are happy with. There are a lot of functionals where we want to pass indices, probably as a compile-time thing. E.g. to specify whether to broadcast or what is an option type or whatever.

insertinterestingnamehere:
Yep. The idiom we were using ought to have worked. If we can get this fixed upstream I'd say let's just go with that. No idea what to do in the mean time though. I guessed `std::array` because it seemed like the thing most likely to get optimized away.

PR #1146, opened 2016-04-15
Avoid harmless decrement of exhausted iterator in topological sort.

insertinterestingnamehere:
This fixes an assertion error in running our test suite with a Debug build built with MSVC 2015.
As near as I can tell, the assertion error was just complaining that the iterator in question was decremented one too many times, regardless of the fact that it was never actually accessed out of bounds.

It'd be nice if we could revise the control flow at some point to make it so that the extra if statement is no longer necessary here, but this at least gets the build working again.


izaid:
This looks fine, thanks @insertinterestingnamehere! The fact that we need this means we probably should refactor the algorithm a bit, but we already knew that...

PR #1157, opened 2016-04-19
Implement the dynd string object as per docs/string-design.md

mwiebe:
This implements an SSO bytestring object with a template parameter for whether there should be a nul-byte after the string or not. dynd::string inherits from sso_bytestring<1>, while dynd::bytes inherits from sso_bytestring<0>. See docs/string-design.md for details and rationale.


izaid:
The progidal son returns...

PR #1163, opened 2016-04-20
[WIP] Added tests for outer on type sequences

izaid:


insertinterestingnamehere:
This is great! I'm intrigued that the outer product for type sequences didn't get hit by the MSVC bug that was effecting out integer sequence outer product.

izaid:
Yep! I am moving us towards better dispatch and templating on types ASAP.

PR #1199, opened 2016-05-12
Fix build on FreeBSD.

insertinterestingnamehere:
Stumbled on this while trying to reproduce an OSX test failure.

On BSD operating systems things like dlopen are
included in libc, and libdl does not exist.
I've verified that this works on FreeBSD and Dragonfly BSD.


izaid:
Awesome, thanks! Wish we had a FreeBSD CI config.

insertinterestingnamehere:
Yep. That'd be nice to have. Short term, I don't know of any way to do that. Long term, that's probably possible with anaconda-build, but if that ever happens, it's still a long ways off.

PR #1201, opened 2016-05-12
[WIP] Making memory blocks virtual classes

izaid:


izaid:
This is the beginnings of giving memory_block_data the same semantics as types and callables.

PR #1224, opened 2016-05-16
Made assign_na_kernel be templated on types

izaid:


izaid:
@insertinterestingnamehere Is this more or less what you're thinking with templating on types (not type IDs)? Example is assign_na_kernel.hpp.

I think this is great, it'll ultimately turn the type IDs into a hidden implementation detail.

insertinterestingnamehere:
This looks great to me too! Thanks for doing https://github.com/libdynd/libdynd/pull/1223 as well.
What are the base typedefs doing? There role in all of this isn't clear to me yet.

izaid:
Yeah, so if it's not clear, the plan is to switch over all kernels / callables to this style. I'm doing them a few at a time so I can see where the gaps are -- comparisons were first so I could get a version of nd::callable::make_all working for kernel templates of the form `KernelType<typename...>`.

The base typedefs are to signal what the base type is of a particular DyND type. This information already exists, but it is only done for type IDs. Somewhere we have a `base_id_of<TypeID>` template that has this same information, but specialized on type IDs. Examples:

- int32_t has ndt::int_kind_type as a base type
- ndt::option_type has ndt::any_kind_type as a base type

Make sense?

insertinterestingnamehere:
Got it. That makes sense. Also +1 to that approach.

PR #1229, opened 2016-05-17
[WIP] "Requires" for kernel predicates

izaid:
This is my attempt to allow more finer-grained control of kernel instantiation. It also switches us off using base types as a template enabler, which means we no longer need to know the DyND base of a DyND type at compile time.

Ultimately, this was very easy. Check out uniform_kernel.hpp. We can do exactly what we want with `std::enable_if_t`. However, to make it look more like the syntax we use there, I added a
`template <bool Value> using Requires = std::enable_if_t<Value>;` template alias. Please let me know if you think we should keep `Requires`, or just use `std::enable_if_t` directly.

Pinging @insertinterestingnamehere @mwiebe

### DIFF
--- a/include/dynd/callables/uniform_callable.hpp
+++ b/include/dynd/callables/uniform_callable.hpp
@@ -14,11 +14,12 @@ namespace dynd {
 namespace nd {
   namespace random {
 
-    template <typename ReturnType, typename GeneratorType, typename Requires = void>
+    template <typename ReturnType, typename GeneratorType, typename Enable = void>
     class uniform_callable;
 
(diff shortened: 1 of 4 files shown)


izaid:
I should probably just drop the `Requires` and stick with `std::enable_if_t` directly.

insertinterestingnamehere:
Yep. That'll be more readable, not because the name is any better, but because it's more easily recognizable.

izaid:
Okay, done. Glad that was easy!

insertinterestingnamehere:
Okay, this looks right overall. Great!

PR #1232, opened 2016-05-18
Append CXXFLAGS for proper overriding (e.g. "-O0 -g").

skrah:
This allows CXXFLAGS in the environment to override the official cmake flags. It is especially useful when dealing with setup.py.


izaid:
@skrah Is this good to go? Should I merge this?

skrah:
@izaid Yes, this should be good.

PR #1241, opened 2016-05-19
Move broadcast error

skrah:
This makes broadcast_error part of libdyndt.


insertinterestingnamehere:
+1 to this approach. The broadcast error is really about incompatible types. The interface should be based around types, not around arrays. Thanks for taking care of this.

izaid:
LGTM, thanks!

PR #1243, opened 2016-05-19
Swapped std::integer_sequence for dynd::integer_sequence

izaid:
This removes our custom integer_sequence (and all associated metaprogramming) from DyND, using only std::integer_sequence where necessary.

Ping @insertinterestingnamehere because I know he's asked about why we have that before.


insertinterestingnamehere:
This is great. It's definitely one of the bigger metaprogramming simplifications we've seen in a long while. Just curious, what happened to the tests for outer with type sequence? I remember seeing you add some a while ago, but I can't find them now.

izaid:
The tests for outer for type_sequence are in test_config.cpp for now.

insertinterestingnamehere:
Great, thanks.

PR #1244, opened 2016-05-19
Refactoring around type_sequence

izaid:


insertinterestingnamehere:
This is great that we're removing the `new_make...` named templates. I tried removing new_make_all a while ago and ran into some trouble. Thanks for tackling this!

PR #1256, opened 2016-05-27
Added new type IDs, removed an old one

izaid:


izaid:
Okay, so this PR breaks the Python tests due to the weird clang bug that appears when we add new type IDs. But, I need to add new type IDs to fix dispatch and the type system. So I'm going to let this through and, hopefully, Python will unbreak itself at the end of the tunnel. ccing @mwiebe @insertinterestingnamehere @skrah

PR #1265, opened 2016-06-02
[WIP] Add flag for building libdyndt only.

skrah:
The problem: It would be good to cut down build times for the dynd-python build when libdynd is in the tree. It is already possible to execute `make libdyndt` manually, but I have trouble finding an elegant way to make that happen inside the dynd-python cmake machinery.

Another way would be to add a `DYND_BUILD_LIBDYND` flag, which is what this patch is doing. Since the tests and benchmarks still depend on libdynd, this is still a bit convoluted.

I'm not sure if I like this patch, I'm mainly hoping for better solutions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,10 @@ if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
     option(DYND_INSTALL_LIB
            "Do installation of the built libdynd library to the CMAKE_INSTALL_PREFIX"
            ON)
+# -DDYND_BUILD_LIBDYND=ON/OFF, whether to build libdynd.
+    option(DYND_BUILD_LIBDYND
+           "Build libdynd."
+           ON)
 # -DDYND_BUILD_TESTS=ON/OFF, whether to build the googletest unit tests.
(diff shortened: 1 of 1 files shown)


izaid:
Does cutting down the libdynd build time help here? I think that is what we should work towards, which is sort of happening, but a bit slowly.

skrah:
Yes, it would help, but the dynd-python Travis CI bottleneck is the inplace build (https://travis-ci.org/libdynd/dynd-python/jobs/134455412), where both libdyndt and libdynd are built twice.

The ndt build should only build libdyndt.  Cmake however is very directory oriented, without the flag in this issue it seems very hard to stop cmake from building libdynd.

izaid:
Gotcha. Yeah, agree with you we should solve this.

insertinterestingnamehere:
Hmm. The best I've been able to find thus far for doing this is using the `EXCLUDE_FROM_ALL` property as in http://stackoverflow.com/a/14911387/1935144.

skrah:
@insertinterestingnamehere I've tried ```EXCLUDE_FROM_ALL``` yesterday without success.  I'm not quite sure, but I think the cmake model really wants everything neatly in its own directory, or we need the same as in setup.py now: Three flags 'all', 'ndt', 'nd' and conditionals everywhere.

PR #1272, opened 2016-06-12
Design doc for multidispatch algorithm

mwiebe:
Leading candidate idea is doing a signature subset construction of a DFA, similar to the algorithm for constructing a DFA from an NFA (https://en.wikipedia.org/wiki/Powerset_construction). Will need to work out a few more examples, and rework the wording to frame the idea more clearly.


mwiebe:
This document still has more to come, but will merge it now to get the code changes in.

datnamer:
Hi, I enjoyed your scipy talk. Since  it sounds like method resolution is dynamic, does that mean it wouldn't be useful in tight inner loops like sampling from a distribution type/struct in jit environment? This is aside from the built in broadcasting which sounds soecially optimizated.

mwiebe:
@datnamer thanks! The multi-dispatch is dynamic, which I think is what's needed for processing dynamic arrays. The decomposition of callables execution into several steps is intended to support these kinds of JIT use cases. The type resolution is a separate step from kernel instantiation, so a JIT compiler interfacing with DyND callables can invoke type resolution as part of JIT compilation, then bake the instantiated kernels tightly into its code.

datnamer:
Sorry for the typos, posted from my phone. 

Ah ok, so the resolution can still happen at compile time and the functions can be inlined? 

Regarding type signatures, is there a way to define new Kinds in datashape? 

Sometimes it makes sense for a general struct pattern, but sometimes i want a library to define a type where the interface isn't in the concrete data, but in the methods defined on the type. Say `Binomial.rand()`  or `agent.move()`

Another "Kinds" usecase involves datastructures like graphs. I understand that the Dynd model to transform arrays  with pure composable functions .  However in some cases it seems  more natural to define iteration interfaces to work with non array datastructures like iterating over the edges of a graph. 

In this case, I want to define a graph interface by the methods it supports, one of these being __iter__. Do you see this sort of thing in the scope of Dynd/datashape, or are you more looking to standard array usecase? 

(I eventually want to do complex things like have agent with probabilistic decision functions in their fields walk through and interact in network structures, do you think that would be possible at some point?).

PR #1283, opened 2016-06-24
Clean up conda recipe

insertinterestingnamehere:
Auto-detect the number of processors and use that to build the
conda package in parallel on non-CI systems.
Stop statically linking libstdc++.
Clear out some dead code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,6 @@ matrix:
       before_script:
         - export CC="gcc-4.9"
         - export CXX="g++-4.9"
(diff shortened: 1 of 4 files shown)


izaid:
Wait, why do we no longer need to statically link libstdc++?

insertinterestingnamehere:
Not statically linking is fine as long as our binaries don't require a libgcc and libstdc++ more recent than the one currently available as a conda package. The libgcc package was updated to version 5.2.0 some time ago.

PR #1316, opened 2017-06-26
MinGW fixes

insertinterestingnamehere:
This includes several fixes needed to get things working on MinGW again.

This includes a fix for a static initialization catastrophe. A possible alternative to our current handling of static initialization that could allow dependencies between startup items would be to initialize everything via a single call (like we used to) that happens when initializing a global variable showing that the library has been initialized (so everything still happens at load time like it does now). That's just an idea to throw around for now though. This change just gets rid of the initialization failure.


insertinterestingnamehere:
Okay, the Linux CI test failures are leftovers from previous issues. I need to deal with those, but don't have time today. It appears that there is one failure from an issue with an old ubuntu package on the travis CI build image ([likely explanation](https://stackoverflow.com/q/37603238/1935144)), a test failure that is currently outstanding in dynd-python, and a previously existing segfault in the test suite.

Provided the OSX build doesn't show anything unusual, I'll leave this PR up for another day or so and then merge to at least get this round of fixes out of the way.

PR #1317, opened 2017-06-27
CI test fixes

insertinterestingnamehere:
WIP: fix various failures in Travis CI script.


insertinterestingnamehere:
Remaining issues:
 - [ ] Segfault specific to the coverage build
 - [ ] Python bindings need to be updated to export the callable type correctly

insertinterestingnamehere:
The new metaprogramming here needs testing/relocation, but I'm merging this now because it finally gets the tests working again. Given the nature of the segfault with gcc that this fixes I consider that a huge win already.

PR #1318, opened 2017-07-02
Fixes for Clang on Windows

insertinterestingnamehere:
I finally found the special sauce to get things working with clang on windows. This fixes a handful of related bugs. I also updated our vendored copy of googletest while working through the various issues that came up. It didn't end up being necessary, but I left it in since it's a positive move.

This worked for me with a version of clang/llvm built from the latest source (with MSVC 2017) as of a day or two ago, though there are a bunch of warnings that still need to be looked at. A very recent release of clang/llvm would likely work too, but I haven't tested that.

The only way thus far that I've actually been able to get the compiler to work with the CMake build system is to use clang-cl through ninja makefiles generated by CMake. Something like

    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -G "Ninja" ..

can be used to get that to work. This is totally experimental, but hopefully it'll make debugging template issues even easier.

### DIFF
--- a/include/dynd/kernels/assignment_kernels.hpp
+++ b/include/dynd/kernels/assignment_kernels.hpp
@@ -11,6 +11,7 @@
 #include <dynd/eval/eval_context.hpp>
 #include <dynd/fpstatus.hpp>
 #include <dynd/kernels/base_kernel.hpp>
+#include <dynd/kernels/base_strided_kernel.hpp>
 #include <dynd/kernels/cuda_launch.hpp>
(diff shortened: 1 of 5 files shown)


insertinterestingnamehere:
I've rolled back the googletest update since it was causing a mysterious linker error with clang on Linux. That update should happen at some point, but it isn't even remotely a priority. It's not strictly needed for anything here.

PR #1329, opened 2017-12-23
Update to latest googletest.

insertinterestingnamehere:
Another pass at https://github.com/libdynd/libdynd/pull/1319.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,21 +15,17 @@ jobs:
   strategy:
     matrix:
       Release:
-        BITS: 64
-        env:
-          BUILD_TYPE=Release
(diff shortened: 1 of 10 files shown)


insertinterestingnamehere:
Hmm. The same linker errors are still there. My best guess is that something in the newer google test making us run up against a compiler bug. https://bugs.llvm.org/show_bug.cgi?id=22825 and https://bugs.llvm.org/show_bug.cgi?id=35027 are examples of similar bugs, though neither of them fully explains the linker errors here.

insertinterestingnamehere:
Works now and fixes https://github.com/libdynd/libdynd/issues/1370. Merging.

PR #1331, opened 2018-01-27
Route stderr to stdout in powershell command.

insertinterestingnamehere:
It appears that at some point conda started writing more of its messages to stderr, which causes false failures when running it from within powershell. This should work around the issues that causes in our appveyor build script.


insertinterestingnamehere:
Errors look unrelated. I'll track those down separately.

PR #1332, opened 2018-01-27
Another pass at fixing appveyor build

insertinterestingnamehere:
Apparently to suppress the false error converting the stream to text is also needed. Credit to https://stackoverflow.com/a/12866669/1935144 for the idea.


insertinterestingnamehere:
Again, the one test failure is not related to this PR.

PR #1343, opened 2018-03-10
Expand Appveyor Builds

insertinterestingnamehere:


insertinterestingnamehere:
Leaving MSVC Debug build commented out due to https://github.com/libdynd/libdynd/issues/1344. That should be addressed at some point, but it doesn't make sense to let it hold up progress here.

PR #1362, opened 2018-08-31
Address anaconda-client upload issues

insertinterestingnamehere:
For some reason uploading the libdynd conda package fails because a piece of setuptools isn't getting found properly. I'm investigating solutions here and will go ahead and merge when that's fixed.


insertinterestingnamehere:
Okay, this *appears* to have worked with the build in question, so I'll try merging and see where to go from there. The failures showing up are just there because I cancelled the unrelated builds.

PR #1363, opened 2018-12-12
Try using Azure Pipelines for CI

insertinterestingnamehere:


insertinterestingnamehere:
Okay, this worked when I ran the workflow on my own repo, but it's not autodetecting the yml config without having it in the main repo... so, I'll put this in and then go from there.

PR #1366, opened 2020-06-02
Make intro compatible with the latest nbviewer.

insertinterestingnamehere:


insertinterestingnamehere:
Actually this appears to just be an issue with the existing GitHub viewing of notebooks, so the solution is to just use nbviewer. Either way, this bumps the format versions so I'll still merge. CI is likely stale and broken right now.

PR #1376, opened 2020-07-16
Compile Fixes

insertinterestingnamehere:


insertinterestingnamehere:
Okay! Things are now working with all versions of clang on Linux, so this is progress!

